PR #3, opened 2021-01-27
Testcoverage - see previous repo

marcrice12:
Fixes the issue with rails forms (scaffolds) and UJS as default - so we can use the javascript file on the application layout (webpacker asset compilation)
Gov UK styles for forms
Validation on the asset
Mocked RSpec tests for the asset
### Guidance to review
prev repo:
https://github.com/guy-roberts/eyfs-reform-spike/pull/14


spec/models/content_asset_spec.rb
```diff
@@ -1,5 +1,30 @@
 require "rails_helper"
 
 RSpec.describe ContentAsset, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  describe "with an attached file" do
+    before(:each) do
+      @content_asset = ContentAsset.new(title: "Title")
```
tonyto, line 6: I'd like to explicitly set the subject for tests where possible.
https://relishapp.com/rspec/rspec-core/v/3-10/docs/subject/explicit-subject

For this spec it's look something like this

```
subject { ContentAsset.new(title: Faker::Lorem.word) }

before(:each) do
  subject.avatar.attach(io: FILE)
  subject.save!
end

it "has a file attached" do
  expect(subject.avatar).to be_attached
end
```

spec/models/content_asset_spec.rb
```diff
@@ -1,5 +1,30 @@
...
+      @content_asset.save!
+    end
+
+    it "has a file attached" do
+      expect(@content_asset.avatar).to be_attached
+    end
+  end
+
+  describe "required attributes" do
+    it "requires a title" do
```
tonyto, line 17: Same as above regarding use of explicit subject

spec/models/content_asset_spec.rb
```diff
@@ -1,5 +1,30 @@
...
+      record = ContentAsset.new
+      record.validate
+      puts "content asset attributes hash: #{record.errors.inspect}"
```
tonyto, line 20: Remove puts

spec/models/content_asset_spec.rb
```diff
@@ -1,5 +1,30 @@
...
+      expect(record.errors[:title]).to include("can't be blank")
+    end
+    it "requires an asset to be uploaded" do
+      record = ContentAsset.new
+      record.validate
+      puts "content asset attributes hash: #{record.errors.inspect}"
```
tonyto, line 26: Remove puts

spec/views/content_assets/edit.html.erb_spec.rb
```diff
@@ -2,9 +2,7 @@
 
 RSpec.describe "content_assets/edit", type: :view do
   before(:each) do
-    @content_asset = assign(:content_asset, ContentAsset.create!(
-                                              title: "MyString",
-                                            ))
+    @content_asset = FactoryBot.create(:content_asset)
```
tonyto, line 5: Can we move these variables into let definitions
https://relishapp.com/rspec/rspec-core/v/2-11/docs/helper-methods/let-and-let

spec/views/content_assets/index.html.erb_spec.rb
```diff
@@ -2,14 +2,10 @@
...
   before(:each) do
-    assign(:content_assets, [
-      ContentAsset.create!(
-        title: "Title",
-      ),
-      ContentAsset.create!(
-        title: "Title",
-      ),
-    ])
+    @content_asset = FactoryBot.create(:content_asset)
```
tonyto, line 5: Can we move these variables into let definitions
https://relishapp.com/rspec/rspec-core/v/2-11/docs/helper-methods/let-and-let

spec/views/content_assets/new.html.erb_spec.rb
```diff
@@ -2,9 +2,7 @@
 
 RSpec.describe "content_assets/new", type: :view do
   before(:each) do
-    assign(:content_asset, ContentAsset.new(
-                             title: "MyString",
-                           ))
+    assign(:content_asset, FactoryBot.build(:content_asset))
```
tonyto, line 5: Can `content_asset` be extracted out into a let definition

spec/views/content_assets/show.html.erb_spec.rb
```diff
@@ -2,9 +2,7 @@
 
 RSpec.describe "content_assets/show", type: :view do
   before(:each) do
-    @content_asset = assign(:content_asset, ContentAsset.create!(
-                                              title: "Title",
-                                            ))
+    @content_asset = FactoryBot.create(:content_asset)
```
tonyto, line 5: Can we move these variables into let definitions
https://relishapp.com/rspec/rspec-core/v/2-11/docs/helper-methods/let-and-let

tonyto (changes requested):

spec/models/content_asset_spec.rb
guy-roberts-dfe: Explicit subjects like this one above are not too hard to read, although I prefer if the code to instantiate it is not too far from where its used, otherwise you're always flipping backwards to see what is being tested.

But let's avoid implicit subjects because its not always obvious how they are instantiated. I think anonymous subjects make tests harder to read.   

Summary: lets use Tony's idea !  But take care not to end up making the tests harder to read.

spec/models/content_asset_spec.rb
guy-roberts-dfe: Yep explicit

spec/views/content_assets/edit.html.erb_spec.rb
guy-roberts-dfe: Marc, I did hit trouble when I tried to do this with ContentPage tests, so let me know if you get stuck.

spec/views/content_assets/index.html.erb_spec.rb
tonyto: Turns out the rendered view requires the instance variable `@content_assets` otherwise it'll fail with a nil object error.
Ignore that request

tonyto (approved):

PR #7, opened 2021-02-01
Better page index

guy-roberts:
### Context
Pages can now have children

The index page for ContentPages shows a nested structure

### Guidance to review
It's mostly changes to the view code of ContentPages.
You could check it out, run it, and create some child pages
It does not order or rank the pages yet


app/controllers/content_pages_controller.rb
```diff
@@ -13,6 +13,11 @@ def show; end
   # GET /content_pages/new
   def new
     @content_page = ContentPage.new
```
tonyto, line 15: Could inline this...

```
@content_page = ContentPage.new(parent_id: params[:parent_id])
```

tonyto (changes requested):
Hey - I ran the branch locally and it it created nested child pages 🎉 

When I POST/PATCH a new page, the redirect doesn't seem to work. It does a 302 back to the edit page. Odd behaviour, not sure if this is within the scope of this PR or another.

One small change but otherwise 👍

app/controllers/content_pages_controller.rb
```diff
@@ -12,12 +12,9 @@ def show; end
 
   # GET /content_pages/new
   def new
-    @content_page = ContentPage.new
     # If the new page is a child, pass through its parent id
-    if params[:parent_id]
-      @content_page.parent_id = params[:parent_id]
-    end
-    @content_page
+    # Pages with a nil parent_id are top_level
```
tonyto, line 16: I don't think this comment is required - it doesn't make the code clearer and It doesn't reason why the code is required.

spec/requests/content_pages_spec.rb
```diff
@@ -13,137 +13,126 @@
...
-
-   it "redirects to the content_pages list" do
-     content_page = ContentPage.create! valid_attributes
-     delete content_page_url(content_page)
-     expect(response).to redirect_to(content_pages_url)
-   end
- end
-
+  # ContentPage. As you add validations to ContentPage, be sure to
+  # adjust the attributes here as well.
```
tonyto: Could remove these comments as the specs have been added

app/controllers/content_pages_controller.rb
guy-roberts: I left the comment in on purpose.  I think it *emphasises* the use of nil to mean top level.  

Sure, people can figure that out, but they might not notice at first that the parent_id is being used for two purposes, to signify top level pages and to associate children with parents.

But I do think that telling a story is more polite, rather than leaving them to reverse engineer it !

Can you tell that I really don't buy this ruby dogma of no comments in the code !

tonyto (approved):

PR #8, opened 2021-02-01
Alt text

marcrice12:
### Context
alt (text) is now an attribute of ContentAsset
migration
validation (required)
view changes
rspec updated


README.md
```diff
@@ -48,7 +48,7 @@ CONTAINER ID   IMAGE                   COMMAND                  CREATED       ST
 
 cancel the running docker process and then run `docker-compose down` (for a full reset on your environment)
 
-### To ssh into the a docker container
+### To connect to the a docker container
 
 `docker exec -i -t eyfs-reform-spike_app_1 sh`
```
tonyto, line 53:
```suggestion
`docker exec -it eyfs-reform-spike_app_1 sh`
```

db/migrate/20210201170927_add_alt_to_content_assets.rb
```diff
@@ -0,0 +1,5 @@
+class AddAltToContentAssets < ActiveRecord::Migration[6.1]
+  def change
+    add_column :content_assets, :alt, :string
```
tonyto, line 3: Should a `null: false` be added to this column?

spec/models/content_asset_spec.rb
```diff
@@ -25,5 +26,9 @@
       subject.validate
       expect(subject.errors[:avatar]).to include("can't be blank")
     end
+    it "requires an alt text for the asset" do
+      subject.validate
```
tonyto, line 30: Could extract `subject.validate` into a `before` block to reduce the repetition on this block of specs

spec/views/content_assets/index.html.erb_spec.rb
```diff
@@ -2,14 +2,14 @@
...
 RSpec.describe "content_assets/index", type: :view do
   before(:each) do
-    @content_asset = FactoryBot.create(:content_asset)
-    @content_asset2 = FactoryBot.create(:content_asset)
-
-    assign(:content_assets, [@content_asset, @content_asset2])
+    @content_assets = [FactoryBot.create(:content_asset), FactoryBot.create(:content_asset)]
   end
 
   it "renders a list of content_assets" do
```
tonyto, line 8: This test might be a false positive because the `FactoryBot` objects created would both contain the same `title` and `alt` text.

Could try using the `Faker` gem to generate text and alt values for the `ContentAsset` Factory object so that they are unique or at least contain different values.

spec/views/content_assets/show.html.erb_spec.rb
```diff
@@ -7,6 +7,7 @@
 
   it "renders attributes in <p>" do
```
tonyto, line 8: How does this test check that the attributes are rendered with a `<p>` tag?

db/migrate/20210201170927_add_alt_to_content_assets.rb
marcrice12: check, had to add default though as well for rubocop

README.md
marcrice12: check

spec/models/content_asset_spec.rb
marcrice12: extraction complete

spec/views/content_assets/index.html.erb_spec.rb
marcrice12: at the moment this test is just testing for a list of (2) content assets with a title and alt text

spec/views/content_assets/show.html.erb_spec.rb
marcrice12: removed <p> from description

spec/views/content_assets/index.html.erb_spec.rb
tonyto: The rendered page will have something like this
```
<div>
  <div> Title </div>
  <div> Sample alt text </div>
</div>
<div>
  <div> Title </div>
  <div> Sample alt text </div>
</div>
```

The test is checking that the page contains the same asset `title` and `alt`.
If the test doesn't require the rendered view to contain two unique assets then the test should not loop through a list of assets.

This test will still be valid if the title and alt appeared on the rendered page 5 times. Which I don't think is the intention

tonyto (dismissed):

tonyto (approved):

PR #21, opened 2021-02-09
/check endpoint for healthcheck pings

tonyto:
### Context

PaaS requires a healthcheck endpoint

Adding a `/check` route that will respond with the following as an example

```
{"status":"OK","version":SHA,"environment":"development"}
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,10 @@
 class ApplicationController < ActionController::Base
   layout :layout_by_resource
 
+  def check
+    render json: { status: "OK", version: ENV["SHA"], environment: Rails.env }, status: :ok
+  end
+
 private
 
   def layout_by_resource
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,6 +10,7 @@ Rails.application.routes.draw do
   end
 
   root "content_pages#index"
+  get "check" => "application#check"
 
   get "/404", to: "errors#not_found", via: :all
   get "/422", to: "errors#unprocessable_entity", via: :all


app/controllers/application_controller.rb
```diff
@@ -1,6 +1,10 @@
 class ApplicationController < ActionController::Base
   layout :layout_by_resource
 
+  def check
+    render json: { status: "OK", version: ENV["SHA"], environment: Rails.env }, status: :ok
+  end
```
guy-roberts, line 6: Does the json include the Rails environment ?  Is there anything in there that would be useful to bad people ?

guy-roberts (commented):
Looks good to me

Just a question, which will not stop the approval, about if the Rails env is shared in the JSON response ?

app/controllers/application_controller.rb
marcrice12: looks like a simple json response object: status string, encrypted version string (SHA), string saying the environment: "production" or "development"

app/controllers/application_controller.rb
```diff
@@ -1,6 +1,10 @@
 class ApplicationController < ActionController::Base
   layout :layout_by_resource
 
+  def check
+    render json: { status: "OK", version: ENV["SHA"], environment: Rails.env }, status: :ok
```
marcrice12, line 5: where does ENV["SHA"] originate?

tonyto: It'll be set when the app is deployed. If it's not set, it'll be `nil`

tonyto:
@guy-roberts `Rails.env` will have one of the follow values `development`, `test, `staging`, `production`

guy-roberts (approved):

PR #32, opened 2021-02-11
Govspeak

marcrice12:
### Context
Govspeak as the default text formatter of our CMS and with a live preview.

The original live preview has branch has been superseded by this one. 

The decision was made to replace Markdown text formatter with Govspeak, as Govspeak offers more accessibility features than standard markdown - which emerged from my review of other gov.uk sites and the kind of accessibility features required to meet the accessibility standards gov.uk subscribes to.

The reason we opted for Markdown initially was that there existed various useful open source projects for Markdown features (not so for Govspeak). For instance i was able to quickly add a JS markdown live preview feature on the editor's page. 

For this PR i was able to retain the live JS preview functionality, with Govspeak, by creating a Govspeak controller which returns html in a JSON response (API) for every Govspeak string it receives. The edit page sends a request to /govspeak on keyup in the textarea - so it's essentially a seamless transition from the previous live preview functionality.

Govspeak replaces Markdown as default text formatter

### Guidance to review

Govspeak is not simply an expanded version of Markdown, it works differently, so some of the dummy copy we have in there wont look formatted correctly
Screenshot 2021-02-11 at 21 28 54

A new line is required before subheadings as shown in above example you can verify this by looking at the govspeak preview site

https://govspeak-preview.herokuapp.com/


app/controllers/govspeak_controller.rb
```diff
@@ -0,0 +1,10 @@
+class GovspeakController < ApplicationController
+  skip_before_action :verify_authenticity_token
```
guy-roberts, line 2: we need to make sure only authenticated users can use it

  before_action :authenticate_user!

db/schema.rb
```diff
@@ -64,6 +64,7 @@
     t.string "subtitle"
     t.integer "parent_id"
     t.integer "position"
+    t.string "published_status"
     t.index ["position", "parent_id"], name: "index_content_pages_on_position_and_parent_id", unique: true
```
guy-roberts, line 68: I removed published_status,  lets not add it back until the story for status arrives

spec/controllers/govspeak_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        expect(JSON.parse(response.body)).to have_key("html")
+      end
+
+      it "returns valid html" do
+        post "show", params: valid_params
+        expect(JSON.parse(response.body)["html"]).to include('<h2 id="govspeak-is-magic">Govspeak is magic!</h2>')
+      end
+    end
+  end
+end
```
guy-roberts, line 26: Add a test to show you have to be logged in to POST content

Otherwise the Russians will call the controller and edit a government web site

guy-roberts (changes requested):
Hi, I tried this out locally.  

The new controller needs to require authentication, otherwise anybody could POST to a government web site

There is a lot of markdown output in the logs, it will flood the production logs, anyway to turn it off ?  The output is more verbose than other actions because he markdown param is long.

app/controllers/govspeak_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+
+  require "govspeak"
+
+  def show
+    doc = Govspeak::Document.new params[:input]
+    render json: { html: doc.to_html }
+  end
+end
```
guy-roberts, line 10: Can you stop the markdown param appearing in the logs ?

Because it will make it hard to see anything else

This shows how

https://stackoverflow.com/questions/63147598/disable-certain-parameters-from-being-logged-in-ruby-on-rails

guy-roberts (changes requested):

db/schema.rb
marcrice12: if i remove that ill do it with a migration

app/controllers/govspeak_controller.rb
marcrice12: check

app/controllers/govspeak_controller.rb
marcrice12: ive added     Rails.logger.silence  to that method

db/migrate/20210212160639_remove_published_status_from_content_page.rb
```diff
@@ -0,0 +1,5 @@
+class RemovePublishedStatusFromContentPage < ActiveRecord::Migration[6.1]
```
tonyto, line 1: As this is not in `main` I don't think you need this migration.

You might need to rebase with `main`, drop db, create db and run the existing migrations from the start to get your environment realigned.

tonyto: /cc @guy-roberts

spec/requests/govspeak_request_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+# RSpec.describe "Govspeak", type: :request do
```
tonyto, line 3: Why is this spec commented out?

yarn.lock
```diff
@@ -3982,6 +3982,11 @@ jest-worker@^26.5.0:
     merge-stream "^2.0.0"
     supports-color "^7.0.0"
 
+jquery@^3.5.1:
```
tonyto, line 3985: Yikes - I feel like jQuery is a bit heavy for what we're using it for.

marcrice12: i think theres probably enough js pipelined to justify it, trying to put a simple structure in place for our js code

tonyto (dismissed):
Some things to fix and discuss at a later date. 
LGTM

guy-roberts (approved):

guy-roberts (approved):

PR #37, opened 2021-02-14
Fixes for sprint 2 test fails

guy-roberts:
### Context
Three minor fixes for issues found by Darren when he ran his tests

The Landing page was missing Chevrons, 
The first three sections were not in order, 
The db constraint of position being unique had not been removed when it was changed to be a compound key.

### Guidance to review
1. Create two children of two parent pages,  give both children the same position, say 342
It should allow both to be created without an exception being shown
2. Chevrons should appear on the Landing page
3. The order of sections on the landing page should look like this
Screenshot 2021-02-14 at 20 23 25

The landing page should have chevrons to the right of each of the seven areas
Screenshot 2021-02-14 at 20 17 13


app/webpacker/packs/application.js
```diff
@@ -1,6 +1,9 @@
 require.context('govuk-frontend/govuk/assets');
 
 import '../styles/application.scss';
+
+const images = require.context('../images', true)
```
tonyto, line 5: I'm not sure I understand why this is set as a `const` - could you add some context?

tonyto (commented):
There's some images added to webpacker but I can't see them referenced. Are they required?

app/webpacker/packs/application.js
guy-roberts: Fair enough.
I cut and pasted and did not come back to tidy it up.
Gone soon

guy-roberts: About the assets that are not used, I deliberately copied them all because I expected they would be used in some part of the prototype code I have not seen yet.  And because they are tiny, 196 bytes

The print icons are needed, at least one

tonyto (approved):

PR #39, opened 2021-02-16
Hfeyp 51  mobile nav

marcrice12:
### Context

https://dfedigital.atlassian.net/browse/HFEYP-51

Mobile nav based on: 
https://eyfs-framework.herokuapp.com/main/Round2Version1/communication-and-language/conversations#
Screenshot 2021-02-15 at 19 03 13


app/helpers/content_helper.rb
```diff
@@ -2,4 +2,8 @@ module ContentHelper
   def path_for_this_page(page)
     "#{content_url}/#{page.slug}"
   end
+
+  def has_active_css(href)
+    "app-mobile-nav__subnav-item--current" if current_page?(href)
+  end
```
guy-roberts, line 8: Yesterday I added a function to do this;

```
  def is_selected(page)
    page.slug == params[:slug]
  end

It is already in main, can you use this (its called is_selected) ? Rather than duplicate it ? Then put the css in the erb file, so that its close to the html. And in the html, use is_selected to conditionally use the css.

I just think its easier to follow if the css is in the erb, not the helper

File: db/seeds.rb [hunk 2]
```diff
@@ -119,4 +119,46 @@
   counting_page.parent_id = ContentPage.find_by(title: "Maths").id
   counting_page.markdown = markdown_for_counting
   counting_page.save!
+
```
guy-roberts, line 122: Please don't add to the seeds.rb,  we changed it with real content today, so this is not needed

app/helpers/content_helper.rb
marcrice12: ok ive switched to using your function, and added erb conditional rendering for class but renamed the method is_current? - ? is ruby convention for boolean and the css classes we need for this use current in their name.

db/seeds.rb
marcrice12: this PR predates the merged seeds PR thats why it's in here. But i've been able to merge main to this and remove the seeds i needed at add at the time..

guy-roberts (approved):

PR #40, opened 2021-02-17
Hfeyp 45  insert uploaded images

marcrice12:
### Context
https://dfedigital.atlassian.net/jira/software/projects/HFEYP/boards/83?selectedIssue=HFEYP-45

### Guidance to review
Screenshot 2021-02-16 at 18 46 12
Screenshot 2021-02-16 at 18 45 52


guy-roberts (commented):
Just thinking, does the link need to be in an editable input ? They might accidentally spoil it. Could it be in a div or something ? Read only.

guy-roberts (commented):
In fact, how about wrapping the link in the full markdown, so that they don't have to figure out the []() syntax, with the alt text ?

Then once its pasted in the page, they can edit the alt text.

So the string they would copy would look like this, 

```
![My fish](http://localhost:3000/rails/active_storage/blobs/redirect/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBCZz09IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--0aeed6cafe7e55931b171d2ae36147bc5244fe69/Screenshot%202021-02-18%20at%2010.36.27.png)

The alt-text could be in there

File: app/webpacker/custom/admin.js [hunk 1]
```diff
@@ -9,4 +9,9 @@ export function createPreview(el, target) {
 
 }
 
-
+export function copyToClipboard() {
```
guy-roberts, line 12: Could you put a comment in here to remind people where the JS is used ?

app/webpacker/styles/application.scss
```diff
@@ -21,6 +21,7 @@ $govuk-image-url-function: frontend-image-url;
 @import "layout-eyfs/landing-page.scss";
 
 @import "_content-pages-edit.scss";
+@import "_content-assets-edit.scss";
 
 #content_asset_avatar, #content-asset-wrapper {
   padding-top: govuk-spacing(6);
```
guy-roberts, line 27: I thought we had removed mentions of avatar ?

Its not an avatar, its an uploaded image

Please can you rename it ?

guy-roberts (commented):
The text copied to buffer should include the full markdown syntax and the alt-text, everything needed to see the image in the markdown.

The editors can edit the alt-text once its in the page if they like

This will save hours of wasted time as they ask about the correct  syntax

Also, the input box should be a span or something, not editable

timallen-and:
Agreed that the text box should not be editable or a div and that variables should be named for their purpose.
Disagree with the whole markdown as that will be covered elsewhere.

tonyto (approved):

app/webpacker/styles/application.scss
marcrice12: removed unused css

PR #49, opened 2021-02-19
Setup pre-prod deployment job

tonyto:
### Context

Deploy app to pre-prod environment from the `prod` branch

Setup job for deployment.
Added some documentation - still a WIP

### Guidance to review

Does anything look off?


guy-roberts (approved):
Looks good to me,  though I don't really know how these declarations are used.

Just curious, does the whole preproduction machine get destroyed and rebuilt each time a version is deployed ? 

Does this mean that the logs and database contents will be dropped ? 

I'm interested to know if the plan is to deploy to production through github too.

Or will it production deployments be triggered by people ?

tonyto:
@guy-roberts Deployments for production will be run through github workflows. They will be manually triggered by a human.

Each deployment will create a new docker image and the infrastructure is updated with that image to run from.

Logs are preserved.

This is the start command that gets run every time a new deployment is made 

```
bundle exec rake db:prepare && bundle exec rake db:seed && bundle exec rails s  -b 0.0.0.0
```

The database does not get dropped, but migrations are run and the db is seeded before starting. The contents of the DB will depend on how the content is created. 
I believe atm, seeding the contents is a destructive process.

PR #60, opened 2021-02-22
Dockerfile for dev environment

tonyto:
### Context

Creating a `Dockerfile.dev` that is referenced in `docker-compose.yml` for local development purposes.
This leaves the original `Dockerfile` for deployment purposes.

Sets required env variables for  local development with docker.


marcrice12:
tested locally, works

marcrice12 (approved):
works locally on dev

PR #62, opened 2021-02-22
HFEYP-183 - Styles for the HTML that GovSpeak produces

guy-roberts:
### Context
The aim is to style the html produced from Markdown. Until now, it's only had styling for h1, h2 and p.   
This PR adds a set of SCSS files (from https://github.com/DFE-Digital/ecf-engage-and-learn) which apply govuk-frontend styles via a parent selector.
This parent class, .gem-c-govspeak is used in the layout for content (content.html.erb) and in the edit preview for pages, so that all of these styles now work.

The .nhs-card classes have been renamed .eyfs-card
CMS styles have been gathered into one file, _cms.scss

### Guidance to review
Run the app and compare pages with those in the prototype app, https://eyfs-framework.herokuapp.com/main/ContentRound2Version1/communication-and-language/communication-and-language-index

This picture shows our code on the left, and the prototype code on the right

image


dmallett1 (commented):
The '-card' reference appears to be from nhs/content design and possibly does not add meaning or benefit within the html.  'eyfs' appears sufficient.

app/webpacker/styles/_cms.scss
```diff
@@ -0,0 +1,39 @@
...
+  margin-left: 4em;
+}
+
+#markdown-editor {
+  height: 350px;
+}
+
+.copy-to-clipboard {
+  margin-bottom: 20px;
+  background: #f3f2f1;
```
tonyto, line 27: Should this be a gov-uk colour variable? https://design-system.service.gov.uk/styles/colour/

guy-roberts: Yes, it should

marcrice12 (approved):
looks ok

PR #64, opened 2021-02-23
36 logout button

marcrice12:
### Context
https://dfedigital.atlassian.net/jira/software/projects/HFEYP/boards/83?selectedIssue=HFEYP-36

1. logout button for the admin area of the site (header)
2. 30 min inactivity session timeout


tonyto (approved):

guy-roberts (approved):

guy-roberts (commented):
I changed the devise config to 
```
config.timeout_in = 1.minutes

then waited a couple of minutes

I could still edit and save pages

When it does work, will it clear the current view and go back to the login prompt or will it wait until another request is made ?

**Review by guy-roberts — CHANGES_REQUESTED** [review]
The user model is not :timeoutable,  so the time out will not happen

**Review by guy-roberts — APPROVED** [review]

File: spec/features/admin_home_page_spec.rb [hunk 1]
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
+
+RSpec.feature "Admin Homepage", type: :feature do
```
tonyto, line 3: Is it worth asserting for the `current_user` for these specs?

PR #68, opened 2021-02-24
Rake task to destroy user

tonyto:
### Context

This PR creates a Rake task that can be run to destroy a user.
The task take an email as an argument.

Added `act_as_paranoid` gem for soft deletions

### DIFF
--- a/lib/tasks/sys_admin_tasks.rake
+++ b/lib/tasks/sys_admin_tasks.rake
@@ -1,0 +1,7 @@
+desc "Tasks triggered by sys admins"
+namespace :sys_admin_tasks do
+  desc "destroy user"
+  task :destroy_user, [:email] => :environment do |_, args|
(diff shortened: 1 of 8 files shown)


lib/tasks/sys_admin_tasks.rake
```diff
@@ -0,0 +1,7 @@
+desc "Tasks triggered by sys admins"
+namespace :sys_admin_tasks do
+  desc "destroy user"
+  task :destroy_user, [:email] => :environment do |_, args|
+    User.find_by(email: args.email).destroy!
+  end
```
guy-roberts, line 6: If they have the wrong email, it will show then an exception.  If this was going to be used by a non ruby programmer, it might be nicer to catch the exception and say "Email not found" on the console

A nice to have

guy-roberts (dismissed):

PR #70, opened 2021-02-25
HFEYP-202 - Next and Previous buttons

guy-roberts:
### Context 
Pages now have Next and Previous navigation buttons (but not the landing page)

Added two methods to the ContentPage model, next_page and previous_page
Added a class variable holding the ordering of pages, which is rebuilt whenever a page is edited
This will help avoid lots of queries to the database every time the next and previous buttons need to be rendered

https://dfedigital.atlassian.net/browse/HFEYP-202

### Guidance to review
Compare the Next and Previous buttons on a page
Check that they work in the correct sense, a depth first search. Clicking 'Previous' on the first page should take you to the last, and vice versa
I've added rspecs


spec/models/content_page_spec.rb
```diff
@@ -43,4 +47,49 @@
     expect(saved_page.slug.count("-")).to be(3)
     expect(saved_page.title.count("-")).to be(2)
   end
+
+  # The order is a depth first search
+  context "Next and Previous pages" do
```
tonyto, line 52: It would be nice to see this test split into two, one block to test the Next unit and another to test the Previous unit.
By testing the methods in isolation leads to less coupling down the line

spec/models/content_page_spec.rb
```diff
@@ -43,4 +47,49 @@
...
+    before(:each) do
+      ContentPage.destroy_all
```
tonyto, line 54: What happens when `ContentPage` is `nil` - It'd be worth getting a test block for that usecase.

tonyto: Also, I think this `before` block is not required because of line 8 in this file

app/models/content_page.rb
```diff
@@ -16,7 +16,53 @@ class ContentPage < ApplicationRecord
 
   before_save :set_slug_from_title
 
+  after_commit do
+    ContentPage.reorder
```
tonyto, line 20: What is the purpose of reordering on an `after_commit` callback? It doesn't seem like ordering is persisted.

guy-roberts: Without reordering, the navigation ignores new pages

The ordering is persisted, in the class variable

spec/models/content_page_spec.rb
guy-roberts: Done

spec/models/content_page_spec.rb
guy-roberts: This is needed so that the earlier pages do not interfere with those created below to test Next and Previous

Really we should start using  the database cleaner gem in tests

app/models/content_page.rb
tonyto: Right, understood. It feels like this behaviour could be moved out into a `ContentPageOrderService` or something. Isolating the ordering logic away from the model.
Then the model can tell the service to break the cache and reorder or ask the service to fetch who my next and previous pages are.

guy-roberts: Honestly, I think acts_as_nested_set is the way, then each page gets a next and previous link.  But for now, it does what it needs to, and I hope we can refactor it later

guy-roberts: Honestly, I think acts_as_nested_set is the way, then each page gets a next and previous link.  But for now, it does what it needs to, and I hope we can refactor it later

tonyto: From what I understand of `acts_as_nested_set` - it requires the model to have 2 fields in the database that are populated. So when a page ordering is changed,  it'll require updates to models the change effects, cascading down the tree. 

I guess the benefit is an order doesn't need to be stored in memory.

guy-roberts: I assumed that acts_as_nested_set would recalculate its idea of next and previous whenever the position changed.  I only skimmed the docs, but thats what I'd guess.

tonyto (approved):

PR #72, opened 2021-02-26
Fixed invalid syntax with delete user workflow

tonyto:


guy-roberts (approved):
Looks good to me

Can't tell what much of the declarative stuff does

PR #73, opened 2021-02-26
179 front-end fixes/updates

marcrice12:
### Context

https://dfedigital.atlassian.net/jira/software/projects/HFEYP/boards/83?selectedIssue=HFEYP-179

These numbered points map to the bulleted list found in the ticket

1. a blue line is removed from content#index 
2. the seeds were updated to have the new link names found here https://eyfs-framework.herokuapp.com/main/Round3Version1/communication-and-language/communication-and-language-index
I also updated the hardcoded names which are found on app/views/layouts/landing_page_layout.html.erb which i assumed would be required to maintain consistency
I also updated the ordering of the seed data to match the ordering on the new prototype 
3. removed the bottom section feedback from views, but kept the code in an unused partial in case we need to bring it back
4. removed pinterest from social icons
3. correct blue used and correct font for social icons

### Guidance to review

1 this is the now removed blue line
Screenshot 2021-02-25 at 17 11 51
without:
Screenshot 2021-02-26 at 17 01 33

2. latest seeds shown on the left and prototype version 3 on the right
Screenshot 2021-02-26 at 16 57 50

4.5. social icons now (4 & 5)
Screenshot 2021-02-26 at 16 54 10


app/views/layouts/landing_page_layout.html.erb
```diff
@@ -167,10 +167,10 @@
               <a class="govuk-link govuk-link--no-visited-state app-mobile-nav__link" href="/main/Round1Version1/maths/counting">Counting</a>
             </li>
             <li class=" app-mobile-nav__subnav-item">
-              <a class="govuk-link govuk-link--no-visited-state app-mobile-nav__link" href="#">Number patterns</a>
+              <a class="govuk-link govuk-link--no-visited-state app-mobile-nav__link" href="#">Numbers in context</a>
             </li>
             <li class=" app-mobile-nav__subnav-item">
-              <a class="govuk-link govuk-link--no-visited-state app-mobile-nav__link" href="#">Spatial reasoning</a>
+              <a class="govuk-link govuk-link--no-visited-state app-mobile-nav__link" href="#">Reasoning</a>
```
guy-roberts, line 173: This whole menu needs to be deleted

It is hidden because of the CSS, but the html is here as a hang over from when it was a place holder for the mobile menu

I think it should be deleted in this PR

guy-roberts (changes requested):

app/views/layouts/landing_page_layout.html.erb
marcrice12: gone

guy-roberts (approved):

PR #77, opened 2021-03-02
Robot - nofollow, noindex and basic auth for content routes

tonyto:
### Context

Stop bots from crawling the website

Adding basic auth to content routes


guy-roberts (approved):
Nice and quick

Maybe after launch it can be changed to do this just for non-production environments

PR #81, opened 2021-03-03
Fix basic auth env

tonyto:
### Context

ENV doesn't like symbols obviously


dmallett1 (approved):
This cannot do any harm given that env is currently broken.
Approved.

PR #85, opened 2021-03-03
Added the audited gem, which will quietly save every change made to the PageContent table

guy-roberts:
### Context
The audited gem will now watch for changes to PageContent and save changes to the new audited table

We don't need to do anything else until somebody wants to know about the page history

This is not versioning, its an audit trail

Added audited to PageContent
Added the audited gem

### Guidance to review
Run the new migration
Run the app and edit a page
Look in the database to see who changed what when
This is what the database table looks like on my dev machine after running the seed data
Screenshot 2021-03-03 at 13 07 43

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,10 +34,10 @@ gem "acts_as_tree"
 
 gem "acts_as_paranoid", "~> 0.7.0"
 
+gem "audited", "~> 4.9"
+
 gem "govspeak"
 gem "htmlentities", "4.3.4"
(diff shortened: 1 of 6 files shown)


tonyto:
The database size could grow very quickly here - it will start showing impacts in space consumption, like, increase in database size and backup size and also time taken for database backup.

Should audit events be stored in another database?

marcrice12 (approved):

tonyto (approved):

PR #86, opened 2021-03-03
Remove seo and subtitle attributes from the ContentPage model

guy-roberts:
### Context
These two attributes are not used

Remove the seo and subtitle from all views, db table, seeds and tests

### Guidance to review
Check that you can still create and edit a ContentPage
Check it saves itself without needing seo and subtitle
Check that the pages still display

### DIFF
--- a/app/controllers/content_pages_controller.rb
+++ b/app/controllers/content_pages_controller.rb
@@ -60,6 +60,6 @@ private
 
   # Only allow a list of trusted parameters through.
   def content_page_params
-    params.require(:content_page).permit(:title, :markdown, :seo, :subtitle, :parent_id, :position)
+    params.require(:content_page).permit(:title, :markdown, :parent_id, :position)
   end
 end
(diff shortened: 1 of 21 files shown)


db/migrate/20210212120340_remove_position_index_from_content_page.rb
```diff
@@ -1,9 +1,15 @@
 class RemovePositionIndexFromContentPage < ActiveRecord::Migration[6.1]
   def up
```
tonyto, line 2: Can you create a new migration to remove the columns rather than change an existing migration.

tonyto (changes requested):

db/migrate/20210212120340_remove_position_index_from_content_page.rb
guy-roberts: My mistake, I meant to do that and must have edited the wrong file

Crikey

guy-roberts: Done now

app/controllers/content_pages_controller.rb
```diff
@@ -60,6 +60,6 @@ def set_content_page
 
   # Only allow a list of trusted parameters through.
   def content_page_params
-    params.require(:content_page).permit(:title, :markdown, :seo, :subtitle, :parent_id, :position)
+    params.require(:content_page).permit(:title, :markdown, :parent_id, :position)
   end
```
guy-roberts, line 64: Gone now

tonyto (approved):

PR #89, opened 2021-03-04
Adding cms app resource

tonyto:


guy-roberts (changes requested):
Whats the bigger picture, story, for this ?

Is it to do with splitting the CMS and public sites ? I can see its setting up a Rails app, for CMS, but don't feel like I understand how it will all work yet.

guy-roberts (approved):

marcrice12 (approved):

PR #90, opened 2021-03-05
Restricting web_app resource to read only binding of postgres

tonyto:
### Context

With two versions of the application deployed on separate instances, one for the CMS and another for the website. They both share the same Postgres database instance but only the CMS application should be able to read and write to the database.

This PR restricts the binding to Postgres to be read only for the website instance. Removing write permissions.
Write permissions would still remain for the CMS instance


guy-roberts (commented):
Good stuff

marcrice12 (approved):

PR #95, opened 2021-03-07
Hfeyp 65 and 118 get help

guy-roberts:
### Context

https://dfedigital.atlassian.net/browse/HFEYP-65

https://dfedigital.atlassian.net/browse/HFEYP-118

The landing page has changed, its "Get Help to improve your practice" section is now dynamic and shows a link called "Overview" pointing to "Get Help to improve your practice" and also links children of that page, ordered by position

### Guidance to review
Check that the landing page renders as above


tunde-anddigital (approved):
checked the page, overview renders fine, child pages show as required

app/helpers/content_helper.rb
```diff
@@ -6,4 +6,8 @@ def path_for_this_page(page)
   def is_current?(page)
     page.slug == params[:slug]
   end
+
+  def find_page_with_title(title)
+    ContentPage.find_by_title(title)
```
tonyto, line 11: I feel like this belongs in the controller. 

Assigning this to a variable and then using it in the view feels like a cleaner way to make a database call to fetch content rather than using a view helper.

app/views/layouts/_show_links_to_children_of_page_with_title.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <% if (featured_page = find_page_with_title(title_of_page_to_render)) %>
+
+    <h2 class="govuk-heading-l"><%= title_of_page_to_render %></h2>
+    <p class="govuk-body"><a class="govuk-link" href="<%= path_for_this_page(featured_page) %>">Overview</a></p>
+
+    <% (featured_page&.children || []).each do |child| %>
+        <p class="govuk-body"><a class="govuk-link" href="<%= path_for_this_page(child) %>"><%= child.title %></a></p>
+    <% end %>
+  <% else %>
+    <p class="govuk-error-summary"><%= "Error: Could not find page called #{title_of_page_to_render}" %></p>
```
tonyto, line 11: Where would this error appear on the page? Would it be better to not show an error if it appears half way down the page?

app/helpers/content_helper.rb
marcrice12: Agree with tony 100% this needs to be an instance variable in the corresponding controller action. Database queries should not take place in any part of the views.

guy-roberts: Ok will do, though the mobile menu layout view partial makes lots of database calls too.

guy-roberts: The longer term answer is to introduce a featured page attribute, 'Get help with your practice' would have an attribute set, featured_page: 1,  and then the view would just render that featured page.  Rather than hard coding the page title in the view.

marcrice12: > Ok will do, though the mobile menu layout view partial makes lots of database calls too.

good spot, we need to refactor that at some point

guy-roberts: Screenshot 2021-03-08 at 15 00 06

Here you go

app/views/layouts/_show_links_to_children_of_page_with_title.html.erb
guy-roberts: I've added an error higher up now.

guy-roberts: Screenshot 2021-03-08 at 15 00 06
Here you go. Really this will only happen if the editors change the name of the 'Get Help Page'.

tonyto (approved):
Nice work

PR #103, opened 2021-03-09
Add DNS route mapping to prod deployment

tonyto:
### Context

On deploy to production the DNS route is lost, meaning the domain 404s

Remaps the DNS route after deploying to production


timallen-and (approved):
Looks good

PR #104, opened 2021-03-09
Prod delete user

tonyto:
### Context

Restrict delete user task to prod only

Updated documentation


marcrice12 (approved):
all good, pr removes unneeded installs for non prod envs

PR #112, opened 2021-03-11
Removed autocomplete removed from login input fields

tonyto:
### Context

Autocomplete should be set to off on Login page

This PR removes the autocomplete attributes from signin input fields


guy-roberts (changes requested):
There are 14 other mentions of autocomplete in forms.

Do they need to be removed too ? Or just the login ? Dunno

guy-roberts (approved):

PR #116, opened 2021-03-15
Feature/service object

tunde-anddigital:
### Context
**HFEYP-268**
https://dfedigital.atlassian.net/jira/software/projects/HFEYP/boards/83?selectedIssue=HFEYP-268

At the moment, the live preview of Markdown only works because the controller disables verify_authenticity_token ...

Remove the skipping of authentication when editing markdown, add x-crsf token in header before post request

### Guidance to review

- Check live preview still works when editing markdown
- Verify authentication is not skipped??


app/controllers/content_controller.rb
```diff
@@ -21,8 +19,7 @@ def show
       end
     end
 
-    doc = Govspeak::Document.new(@page.markdown, sanitize: true, allowed_elements: ContentController::ALLOWED_TAGS)
-    @markdown = doc.to_html
+    @markdown = MarkdownToHtml.new.translate_markdown(@page.markdown)
```
marcrice12, line 22: Govspeak is not markdown, it doesnt work the same so we need to avoid using the term as its misleading, so no new classes or variables should be added which use the term markdown, can you rename it to Govspeak please

app/services/markdown_to_html.rb
```diff
@@ -0,0 +1,8 @@
+class MarkdownToHtml
```
marcrice12, line 1: theres no rspec tests to cover this class, it needs to be well tested as its an important part of the site

app/views/content/show.html.erb
```diff
@@ -4,4 +4,4 @@
 <h1 class="govuk-heading-l">
   <%= @page.title %>
 </h1>
-<%= sanitize(@markdown, tags: ContentController::ALLOWED_TAGS ) %>
+<%= sanitize(@markdown, tags: MarkdownToHtml::ALLOWED_TAGS ) %>
```
marcrice12, line 7: again use of markdown in the class name, needs to refactored out

app/controllers/content_pages_controller.rb
```diff
@@ -22,8 +22,8 @@ def new
 
   # GET /content_pages/1/edit
   def edit
-    doc = Govspeak::Document.new @content_page.markdown, sanitize: true, allowed_elements: ContentController::ALLOWED_TAGS
-    @md = doc.to_html
+    @md = MarkdownToHtml.new.translate_markdown(@content_page.markdown)
```
marcrice12, line 25: instead of chaining the new method to translate_markdown like this, can you refactor the class so translate_markdown is a class method

marcrice12 (changes requested):
changes requested in comments

guy-roberts (approved):

PR #124, opened 2021-03-19
Add validations for content assets

ebrett:
### Context
Adds validations for content assets

* content type
* extension

### Guidance to review
Does not include antivirus updates, still under investigation


app/models/content_asset.rb
```diff
@@ -1,6 +1,24 @@
 class ContentAsset < ApplicationRecord
+  VALID_FILE_EXTENSIONS = %w[.pdf .doc .docx .xls .xlsx .jpeg .jpg .png].freeze
+  VALID_CONTENT_TYPE = %w[
```
guy-roberts, line 3: If somebody tries to upload picture.JPG, will it let them ?

config/locales/en.yml
```diff
@@ -31,3 +31,10 @@
 
 en:
   hello: "Hello world"
+  activerecord:
```
guy-roberts, line 34: Do we need Hello World ?

guy-roberts (approved):

config/locales/en.yml
ebrett: I would think not

PR #126, opened 2021-03-22
Small update for content type image/png

ebrett:
### Context

Missed content type image/png
Allow for uppercase extension (should be valid)


guy-roberts (approved):
Looks good to me

PR #142, opened 2021-03-25
282 - Next and Previous pages now stored in the ContentPage record

guy-roberts:
### Context
https://dfedigital.atlassian.net/browse/HFEYP-282

Currently the next and previous pages are held in memory in a collection. The trouble is that we use two apps, so the second app does not get to update its copy of this ordering when a change is made

Note: after deploying this to production, someone will need to open a rails console and run

```
ContentPage::reorder
```

The ContentPage table now has next_id and previous_id columns
After a page is created, a class method called ContentPage::reorder is called to update all prev/next values and save the changes

### Guidance to review
On the cms site, check that Prev/Next work
Add a page and check again
Check on the non-cms site that the new ordering is shown

Do this again after changing the position of a child page

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 Department for Education - Digital
+Copyright (c) 2021 Department for Education - Digital
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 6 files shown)


ebrett (approved):
looks good to me!

PR #147, opened 2021-03-29
Updates for cookie policy page

ebrett:
### Context

Updates for cookie page needed for 12/4

    Updates for cookie policy page

    * rename session from boilerplate
    * offer to return to home page after setting cookie preference
    * update wording on page, add essential and optional cookies

### DIFF
--- a/app/views/settings/cookie-policy.html.erb
+++ b/app/views/settings/cookie-policy.html.erb
@@ -6,7 +6,7 @@
           <div class="govuk-success-summary" data-qa="cookie-consent-success-message" aria-labelledby="success-message" tabindex="-1" role="alert" data-cookie-confirmation="true">
             <div class="govuk-success-summary__title">
               <h2 class="govuk-heading-m" id="success-message">
-              <%= notice %>
+              <%= notice.html_safe %>
               </h2>
             </div>
           </div>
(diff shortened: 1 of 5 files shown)


app/views/settings/cookie-policy.html.erb
```diff
@@ -76,6 +81,11 @@
               <td class="govuk-table__cell">Checks if you’ve visited Help for early years providers before. This helps us count how many people visit our site.</td>
               <td class="govuk-table__cell">24 hours</td>
             </tr>
+            <tr class="govuk-table__row">
+              <td class="govuk-table__cell">_gat_UA-147538067-8</td>
+              <td class="govuk-table__cell">Checks if you’ve used Help for early years providers before. This helps us count how many people visit our site.</td>
```
guy-roberts, line 86: The GA token looks a bit odd in the code, but then again its probably not going to change, and when it does the developer can find it here

guy-roberts (approved):

app/views/settings/cookie-policy.html.erb
ebrett: Yea, I looked at some other sites and it appears to be what others do.  If you can, please check from you machine and let me know it is the correct number.

guy-roberts (approved):

PR #152, opened 2021-04-01
288 - Login for production content 

guy-roberts:
### Context
Before launch, a small group of users will check the content.   This PR provides a way to turn Basic Auth on, using an environment variable called USE_BASIC_AUTH.

https://dfedigital.atlassian.net/browse/HFEYP-288

If USE_BASIC_AUTH is set, the user will be challenged for their credentials.  Their email address and password are checked against the Devise credentials in the Users database table. 

So we will need to temporarily add these users to production in the same way that CMS users are added.

At the moment, these users would in fact be able to use the CMS to change content. We can add authorisation using cancancan to prevent this.

### Guidance to review
Set the env var to anything and start rails.  Check that all requests for content are met with a Basic Auth dialog.
When credentials are provided, the users should be able to carry on as normal.

_**NB, one request spec needs to be corrected before this PR can be merged,  "With the environment variable USE_BASIC_AUTH set, does not show a basic auth dialog if the user is already logged in with Basic Auth"**_


spec/requests/content_request_spec.rb
```diff
@@ -32,4 +36,35 @@
...
+        ENV["USE_BASIC_AUTH"] = "true"
+
+        get "/"
+        expect(response.status).to eq(401)
+
+        ENV["USE_BASIC_AUTH"] = cached_env_var
+      end
+
+      # This test needs to pass before the PR is accepted
+      xit "does not show a basic auth dialog if the user is already logged in with Basic Auth" do
```
ebrett, line 56: Does this need to be updated to 'it'?

ebrett (changes requested):
Not sure the best way, but we can get all specs to pass if the USE_BASIC_AUTH env var is set?

spec/requests/content_request_spec.rb
guy-roberts: Yes this test needs to be fixed. I am having trouble passing the HTTP_AUTHORIZATION token

And I had not thought to set USE_BASIC_AUTH for all the tests.

Thanks for spotting that, will fix

ebrett (approved):

PR #159, opened 2021-04-07
288 using ordinary auth

guy-roberts:
### Context
288 - adding auth for the 'service users', the first people to see the site, before its open to the public

Another ticket will stop users editing content https://dfedigital.atlassian.net/browse/HFEYP-320

Replace the Basic Auth login form with the ordinary one

### Guidance to review
If the environment variable AUTH_ON_EVERYTHING is set, people will have to log in to see any content at all


ebrett (approved):
Much cleaner implementation. I tested it without cookies allowed and it all still worked.  We need the authorisation part before we release username/passwords though.

PR #173, opened 2021-04-12
Update content security policy

ebrett:
### Context

https://dfedigital.atlassian.net/browse/HFEYP-236

No longer using govuk gem (turns out it's built in to Rails now)

Cherry picked the Google and YouTube bits from the gem and added
DfE specific domains and S3 Domains

### Guidance to review

Test by adding images and using those images in the same environment.

"Cross environment" image posting should work, but should be
discouraged.

If images from any other domains are needed, these domains will need to
be added.


guy-roberts (approved):
A bit hard to tell if its bullet proof, lets deploy and find out !

PR #177, opened 2021-04-14
Add antivirus scanner

ebrett:
### Context

Scans for virus on file upload via ClamAV

Initially proposed to add as a service in the docker compose file.  Instead adding the clamav package to the rails app docker file appears to be sufficient.

### Guidance to review

Will do a quick test on dev.  However, may need to set up cron to refresh the virus database.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,6 +37,8 @@ RUN apk --no-cache add --update \
   postgresql-dev \
   shared-mime-info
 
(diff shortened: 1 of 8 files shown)


guy-roberts (approved):
Probably need to use vmstat to check how much memory it uses when idle, just in case it hogs a lot of resources - it probably won't, but just in case

PR #179, opened 2021-04-15
320 authorisation

guy-roberts:
### Context
Adds authorisation to ContentAssets (ContentPages is already done)

This is the same as the previous PR with this title, except that ContentAssets are now subject to Authorisation as well as ContentPages.

The only classes to have changed since the last PR (approved) are the ContentAssetController and ContentAssetPolicy

Only users with a role of 'editor' will now be able to create or edit ContentAssets

### Guidance to review
I tried to add some tests, but it was taking a long time and we need this feature in production now, so I will finish the tests on another day


ebrett (approved):
Yes, assuming when you rebase/merge with main only the new stuff will remain :)

PR #184, opened 2021-04-19
Turn off AUTH_ON_EVERYTHING - for load testing

guy-roberts:
### Context
For load testing, we need to let the callers get to the pages without needing to log in

Temporarily remove the environment variable called AUTH_ON_EVERYTHING

### Guidance to review
Should not have any effect on dev or test


guy-roberts (commented):
For load testing

PR #209, opened 2021-04-27
367/feature test

dmallett1:


e2etests/features/support/common.rb
```diff
@@ -142,32 +144,32 @@ def tab_click(obj, tab_cnt)
 def check_value(actual, expected)
   if actual != expected
     puts "FAIL Expected: '#{actual}'  Actual: '#{expected}'"
-    @e = "e"
+    @excep = "e"
   end
```
guy-roberts, line 148: When is @excep used ?  If its just the return value of the check_value function then just "e" will be enough, but then, thats always the same, so no point checking it

e2etests/features/support/common.rb
```diff
@@ -142,32 +144,32 @@ def tab_click(obj, tab_cnt)
...
 end
 
 def check_item(pos, desc)
-  if @menu[pos-1] != desc
-    puts "FAIL Expected: '#{desc}' at position '#{pos}' Actual: '#{@menu[pos-1]}'"
-    @e = "e"
+  if @menu[pos - 1] != desc
+    puts "FAIL Expected: '#{desc}' at position '#{pos}' Actual: '#{@menu[pos - 1]}'"
+    @excep = "e"
   end
```
guy-roberts, line 162: Same, where is this @excep used ?

guy-roberts (approved):

PR #241, opened 2021-05-18
docker related seed

dmallett1:

### DIFF
--- a/entrypoints/docker-entrypoint.sh
+++ b/entrypoints/docker-entrypoint.sh
@@ -7,7 +7,5 @@ if [ -f tmp/pids/server.pid ]; then
   rm tmp/pids/server.pid
 fi
 
-bundle exec rake db:prepare db:seed
-
 # Start the application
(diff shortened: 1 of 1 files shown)


guy-roberts (commented):
I guess there is one way to find out, by deploying it to the test env

dmallett1 (commented):
I think I should have left db:prepare in

guy-roberts (approved):

dmallett1:
Merge

ebrett:
Sorry, just getting to this now -- I think the entry points are only used for `docker-compose` which is only used for local development, not for cloud foundry / terraform deployment

dmallett1:
Actually there were 3 seed points and the docker_entry.sh change has stopped deployments running seeds to test env.

The problem is deployment.yml runs docker which runs docker_entry.sh (the offending script) for the Test env.  There is also an app.json in addition to the variables.tf

PR #242, opened 2021-05-18
Updates for antivirus server

ebrett:
### Context
https://dfedigital.atlassian.net/browse/HFEYP-233

* use the production server by default
* use non-rest only for testing purposes
* add instructions to README
* remove from docker-compose (greatly increases time)


config/initializers/ratonvirus.rb
```diff
@@ -1,9 +1,9 @@
 Ratonvirus.configure do |config|
-  if Rails.env.production?
-    config.scanner = :resty
-    config.storage = :active_storage
-  else
+  if Rails.env.test?
```
dmallett1, line 2: Discussed last week re Resty being available to test in Test env, as testing 'eicar' is different code.  This does not appear to have changed.

ebrett: this is the test environment for automated tests run in rails, nothing to do with the completely separate space called 'test'

dmallett1: My mistake no name change required.  So we can now run this is test space, which is great.

README.md
```diff
@@ -236,3 +236,5 @@ https://github.com/niilo/clamav-rest
 
 This will run a clamav-rest server at http://localhost:9000
 
+Set RESTY_SERVICE_URL to "http://localhost:9000/scan"
+
```
guy-roberts, line 240: Maybe mention that these new env vars need to be set in production ?   

RESTY_USERNAME and RESTY_PASSWORD and PROXY

guy-roberts (approved):

README.md
ebrett: At the moment they don't, but will add a note

config/initializers/ratonvirus.rb
ebrett: The server is currently open to the public, by design

The server is only being used to receive and file and send backs a message on whether that particular file is infected or not.  It does not change any of the existing mechanisms, so files remain in their respective spaces

guy-roberts (approved):

PR #247, opened 2021-05-20
I should have removed this ul before now

guy-roberts:
### Context
https://dfedigital.atlassian.net/browse/HFEYP-400

Also https://dfedigital.atlassian.net/browse/HFEYP-399
 
Tpo fix 400, I removed a ul tag that should have been removed in an earlier PR

The cause of 399 is probably the the aria tag mentioned, mobile-menu-nav, should be the id of the mobile menu. 

### Guidance to review
Check that the mobile menu passes the Axe test in that Jira issue, 400

Same for 399.


ebrett:
Just noting that I added that section to match the prototype.  Does the prototype not pass the aXe test?

ebrett (approved):
As I commented - the `<ul>` was added to match the prototype, but looking at it now, it appears to have the `<ul>` inside a `<li>` of the parent `<ul>` -- I am afraid this is going to cause a regression in the accessibility changes we made.  May have to come back this if it does.

PR #256, opened 2021-05-25
Automate axe

guy-roberts:
### Context
Automate some accessibility checks

Added a gem called axe-core-rspec which makes it possible to run Axe checks from rspec

I've added some expectations, 

```
expect(page).to have_text("Pages")
```

Because this PR is not about fixing those accessibility errors yet, the new specs are marked as pending

They can be fixed in another PR

### Guidance to review
Check that all of the specs pass


app/controllers/content_pages_controller.rb
```diff
@@ -23,22 +23,28 @@ def new
...
   def edit
     authorize @content_page, :edit?
-
     @md = GovspeakToHTML.new.translate_markdown(@content_page.markdown)
 
     @content_page
+  rescue Pundit::NotAuthorizedError
+    flash[:notice] = "You don't have permission to edit pages"
+    redirect_to action: "index"
   end
```
guy-roberts, line 33: I had to change this to get the Capybara/Selenium/Chrome tests to work

Otherwise this (expected) exception bubbled up into the test stack and caused it to fail. The previous call to expect raise_error did not catch it.

app/views/content_pages/index.html.erb
```diff
@@ -1,6 +1,6 @@
-<p id="notice"><%= notice %></p>
+<p id="notice" class="govuk-error-message"><%= notice %></p>
 
-<h2 class="govuk-heading-l">Edit pages</h2>
+<h1 class="govuk-heading-l">Edit pages</h1>
```
guy-roberts, line 3: An example of something it found

ebrett (approved):
I am interested in seeing if this has a significant increase in the run time for tests

PR #262, opened 2021-05-26
Explain the clamav-rest server set up in CLAMAV.md

ebrett:
### Context

Explain the `clamav-rest` server set up in CLAMAV.md

* manifest.yml file used to create private servers in eyfs-prod
* default clamav service url points to new server in eyfs-prod


guy-roberts (approved):
My only thought is if some of the URLs could be in env vars

eyfs-clamav-rest-private.apps.internal

But I might be being too fussy

PR #264, opened 2021-05-27
Remove 2px padding from mobile menu

guy-roberts:
### Context
Remove the 2px padding from the left of the mobile menu

### Guidance to review
The lefthand blue border of the selected menu item should now be flush on the left


dmallett1 (approved):
Lets see how it changes.

PR #267, opened 2021-05-28
239/sessions

ebrett:
### Context

https://dfedigital.atlassian.net/browse/HFEYP-239

As a security administrator 

I want users to be able to log in to CMS using a single browser at any one time

So that if they leave a session open on another computer/browser they are immediately logged out of that other session for security reasons. 

#### Acceptance criteria:

 1. Should only be able to log in on one browser at any one time 
 ~2. Should not be able to log in using a browser and the browser’s private/incognito mode.~

Added the `devise-security` gem to enable :session_limitable - when enabled the following message is received

CleanShot 2021-05-28 at 12 49 12@2x

### Guidance to review

Unsure why the 2nd acceptance criteria is there, so reserving that for another PR.  This PR only addresses the first acceptance requirement.


guy-roberts (approved):
I like the extra password validation too

PR #272, opened 2021-06-02
Use errors instead of notices to let the user know that they don't have permission

guy-roberts:
### Context
A fix for the intermittent test fails

Use errors instead of flash notices to tell the user that they don't have sufficient permission

### Guidance to review
The specs should all pass, repeatedly !


spec/features/content_pages_spec.rb
```diff
@@ -55,9 +55,17 @@
 
   scenario "A user with the role of reader should be not be able to edit pages in the CMS" do
     sign_in FactoryBot.create(:user, :reader)
```
ebrett, line 57: Can we say 'not able to update pages' - just to be clear

ebrett (approved):
It really should work the other way -- we should authorise access to the edit page, but this is just as good I think.

PR #275, opened 2021-06-03
Use RACK to block http trace

ebrett:
### Context

https://dfedigital.atlassian.net/browse/HFEYP-241

My attempt at blocking HTTP TRACE.  I found this by googling and finding the implementation here: https://github.com/lockstep/rails_new

It made sense, however unclear how to test it.

### Guidance to review

Still looking for a CURL command to test....


guy-roberts (approved):
Worth a try.  You could test it manually with 

 curl -v -X TRACE http://www.yourserver.com

And I bet there is a way to send a HTTP TRACE request in rspec

PR #284, opened 2021-06-08
Fix the cloud foundry network policy directive

ebrett:
### Context

Documentation was incorrect, should be `cloudfoundry_network_policy` -- also was recommended that we no --reconfigure


dmallett1 (approved):
Let's give it a go as we cannot at present deploy.

PR #285, opened 2021-06-08
Policy definition needs to be explicit

ebrett:
### Context

Fix policy definition


dmallett1 (approved):
I think this is the fix :)

PR #304, opened 2021-06-17
301/add specs

ebrett:
### Context
Just adding some more specs


spec/features/admin/user_admin_page.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+RSpec.feature "User administration", type: :feature do
+  before do
+    create :admin, first_name: "Barbara", last_name: "Gordon", email: "birdsofprey@oracle.com"
+    create :editor, first_name: "Perry", last_name: "White", email: "pwhite@dailyplanet.com"
```
guy-roberts, line 6: You can use  

FactoryBot.create :user, role: 'admin'
FactoryBot.create :user, role: 'editor'

or add a couple of traits to the factory

guy-roberts: That way, if the model gets a new attribute, you only need to update the factory in one place

guy-roberts: A bit like the way parent_page and child_page are done in 

https://github.com/DFE-Digital/early-years-foundation-reform/blob/main/spec/features/content_pages_spec.rb

guy-roberts (approved):

spec/features/admin/user_admin_page.rb
ebrett: I am using factory bot - and rather than traits I'm using aliases.  I think it works better.

guy-roberts: Oh ok

PR #305, opened 2021-06-17
301/small updates

ebrett:
### Context

as stated, some small updates


config/initializers/devise.rb
```diff
@@ -178,7 +178,7 @@
 
   # ==> Configuration for :validatable
   # Range for password length.
-  config.password_length = 6..128
+  config.password_length = 8..128
```
guy-roberts, line 182: The short minimum length does worry me, but even I am tired of hearing my own voice !

db/seeds/create_users.rb
```diff
@@ -1,13 +1,13 @@
...
-    u.role = "editor"
+    u.password = "11abXY!@"
+    u.role = "admin"
   end
 
   User.create! do |u|
     u.email = "guy@home.com"
-    u.password = "guyguy"
+    u.password = "11abXY!@"
     u.role = "editor"
```
guy-roberts, line 11: We need to avoid seeding public servers with passwords that are in the repo

ebrett: agreed, but what's the alternative?

config/initializers/devise.rb
ebrett: yea, they've warn me down as well

PR #308, opened 2021-06-21
469 bug can remove last user admin

guy-roberts:
### Context
https://dfedigital.atlassian.net/browse/HFEYP-469

An admin cannot now assign herself a new role

### Guidance to review
There is a new spec for this.
But to test it manually, add an Admin user then log in.  Try to edit that user and see that the usual radio buttons to change role are replaced by a paragraph of text explaining that an Admin can't change their own role

### DIFF
--- a/app/policies/user_policy.rb
+++ b/app/policies/user_policy.rb
@@ -1,14 +1,14 @@
 class UserPolicy < ApplicationPolicy
   def index?
-    true
+    permissions?
   end
 
   def show?
-    true
+    permissions?
   end
 
   def create?
-    user.admin?
+    permissions?
   end
 
   def update?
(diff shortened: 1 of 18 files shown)


app/policies/user_policy.rb
```diff
@@ -12,7 +12,12 @@ def create?
   end
 
   def update?
-    permissions?
+    # A user with role 'admin' is not allowed to edit their own role
+    if user == record
+      user.role != User::ADMIN
```
ebrett, line 17: you can also use `user.admin?` to avoid the coupling how the role is implemented (i.e. with a string 'admin')

ebrett (approved):

app/policies/user_policy.rb
ebrett: or, in this case, `!user.admin?`

PR #316, opened 2021-06-24
Password update fixed

ebrett:
### Context

[HFEYP-474](https://dfedigital.atlassian.net/browse/HFEYP-474)
[HFEYP-475](https://dfedigital.atlassian.net/browse/HFEYP-475)
[HFEYP-441](https://dfedigital.atlassian.net/browse/HFEYP-441)

Made reset password token unique, cosmetic updates to password fields, added notification on user admin page

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -13,6 +13,7 @@ class User < ApplicationRecord
   # Include default devise modules. Others available are:
   # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
   devise :database_authenticatable,
+         :recoverable,
          :timeoutable,
          :lockable,
          :session_limitable,
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -30,10 +30,8 @@ class User < ApplicationRecord
     if new_password.present?
       self.password = new_password
       self.password_confirmation = new_password_confirmation
-      save
     else
       errors.add(:password, :blank)
-      false
     end
   end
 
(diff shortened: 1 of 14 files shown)


app/models/user.rb
```diff
@@ -29,7 +30,11 @@ def reset_password(new_password, new_password_confirmation)
     if new_password.present?
       self.password = new_password
       self.password_confirmation = new_password_confirmation
-      save!
+      if save
+        true
+      else
+        false
```
guy-roberts, line 36: Just wondering when this will happen, that save fails, and will the user hear about it ?

spec/features/admin/user_admin_page_spec.rb
```diff
@@ -15,16 +15,17 @@
   end
 
   context "with session" do
+    let(:barbara) { User.find_by(email: "birdsofprey@education.gov.uk") }
```
guy-roberts, line 18: Probably easier to assign this user to a variable at the start and hang on to it, but not a big thing

guy-roberts (approved):
Two little things, but looks good to me

app/models/user.rb
ebrett: Yea, not a super fan of this, but more elegant ways tiggered rubocop, so doing this for now (preferred solution would be not to have admin updating passwords, so should go away once we can correct that -- for now, ugly code serves as a warning)

PR #325, opened 2021-07-01
242 - link checker

guy-roberts:
### Context
https://dfedigital.atlassian.net/browse/HFEYP-242 is harder than it sounds

This is not the end of the story for link checking. The results are not easy to interpret because they contain several causes for a broken link

A future PR might remove the actual link checking code might be replaced by a more robust implementation

This PR provides a LinkChecker class that can be used like this

> lc = LinkChecker.new "https://help-for-early-years-providers.education.gov.uk"
> lc.check_page

There is also a new spec, though it is not in the specs directory because it runs very slowly and is not really a test of the main application

Lastly there is a new rake task

> rake link_check:check_live_site

which checks the live site

### Guidance to review
These two commands should work

> rspec app/services/link_checker_spec.rb 

> rake link_check:check_live_site


ebrett (approved):
I love the smell of green code in the morning.....

PR #328, opened 2021-07-02
Brett/release 4

ebrett:
### Context

Update to CHANGELOG for release 4 (no related HFEYP ticket).  Also, a small reversion to the `terraform init` and versions used, as used previously.  Checking to see if this allows build of docker image which is breaking intermittently at the moment.


dmallett1 (approved):
Approved

PR #340, opened 2021-07-14
456 content blocks

guy-roberts:
### Context
https://dfedigital.atlassian.net/browse/HFEYP-184 - Content Blocks

There are also two tickets for specific content blocks, https://dfedigital.atlassian.net/browse/HFEYP-119 (other usefule resources) and https://dfedigital.atlassian.net/browse/HFEYP-459 (accessibility)

These two specific Jira tickets will be done as part of 184

At the moment, the story in the tickets does not follow the actual steps needed to create and use a Content Block.  I have spoken to Bello and I will up date those stories to match the reality of how it works

Model, View, Controller and tests for ContentBlocks.
Users with the Admin role can change ContentBlocks, nobody else can.  Blocks are mentioned in the views using this helper

<%== insert_block('accessibility') %>

In this case its in accessibility_statement.html.erb 

For this to work, a ContentBlock with a name 'accessibility' must exist.  I've added seed data for this.

There are only two blocks used in this PR so far, the one above and other_useful_resources.

If the block is not found, an message is printed on the page to say so, as a last resort

### Guidance to review
Run the seed data in create_content_blocks.rb
Check that the accessibility page looks the same as it does on production
Check that the landing page's section called 'Other useful resources' looks the same

Log in as an admin and try editing one of those blocks by clicking on the new "Blocks' menu on the admin page
Log in as an editor and confirm that the link to the list of blocks is not visible.


app/helpers/content_helper.rb
```diff
@@ -19,4 +19,14 @@ def is_current?(page)
   def is_mobile_menu_section_current?(page)
     params[:slug] == page.slug || params[:section] == page.slug
   end
+
+  def insert_block(block_name)
+    block = ContentBlock.find_by_name(block_name)
+    html_to_use = "Error - block not found"
+    if block
+      html_to_use = ContentBlock.find_by_name(block_name).markdown
+    end
```
ebrett, line 28: What are we doing the find twice? could be something like:

` html_to_use = block ? block.markdown : "Error - block not found"`

ebrett (approved):
LGTM but please hold off on merging until we do a release branch

app/helpers/content_helper.rb
guy-roberts: Thanks !

PR #341, opened 2021-07-15
497 and 501 Improve assets page accessibility, google site verification

ebrett:
### Context
[HFEYP-497](https://dfedigital.atlassian.net/browse/HFEYP-497)
[HFEYP-501](https://dfedigital.atlassian.net/browse/HFEYP-501)

Updated the assets form to use the govuk formbuilder, which should be more accessible (if not, we should investigate why). This added a label and hint to the asset file input.  As such I updated the error message for the file extension to just indicated the invalid extension as the valid extensions are now listed in the hint.

I considered using the govuk form builder for the error summary, which looks much nicer, but it has the same issue with only showing one error per field that we have with users.  It would be nice to build off the govuk form builder's error summary component and update it to show all the errors for the fields.

For 501, the google verification file has been placed in the public folder, so it can be verified by google.

### DIFF
--- a/app/views/content_assets/_form.html.erb
+++ b/app/views/content_assets/_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_with(model: content_asset, local: true) do |form| %>
+<%= form_with(model: content_asset, local: true, builder: GOVUKDesignSystemFormBuilder::FormBuilder) do |form| %>
   <% if content_asset.errors.any? %>
     <div id="error_explanation">
       <h2><%= pluralize(content_asset.errors.count, "error") %> prohibited this asset from being saved:</h2>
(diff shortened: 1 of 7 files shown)


guy-roberts (approved):
Looks good

You could add this automatic Axe check to see how accessible the assets page is

    scenario "The content block index page should not have any accessibility errors" do
      sign_in FactoryBot.create(:user, role: User::ADMIN)
      visit "/cms/assets". # Is this the path ?

      expect(page).to be_axe_clean
    end

ebrett:
That axe spec works well - I ran it on the old page (where it failed) and on the new page (it passed).

guy-roberts (dismissed):

PR #342, opened 2021-07-15
Remove refresh from terraform deploy for test

ebrett:
### Context

Revert the terraform apply command for TEST space - do not use refresh


dmallett1 (approved):
Good, due to 
"The terraform refresh command reads the current settings from all managed remote objects and updates the Terraform state to match.
Warning: This command is deprecated, because its default behavior is unsafe if you have misconfigured credentials for any of your providers. See below for more information and recommended alternatives."

PR #345, opened 2021-07-20
Move notice display, so it shows properly

ebrett:
Previously the notice was rendered haphazardly.
A notice should be displayed on the first response from
it being set in the controller, so should be in the layout.

We have 4 primary layouts right now:

* CMS
* Errors
* Application
* Landing Page

### Guidance to review

As the the cookie banner notice logic is now use in general, I updated the `data-qa` label to just say `success-message` rather than `cookie-banner-success-message`


guy-roberts (commented):
Do any notices need to be removed from other view files ?

guy-roberts (approved):

PR #352, opened 2021-07-21
Only show successful cookie setting message when set from cookies page

ebrett:
### Context
[HFEYP-508](https://dfedigital.atlassian.net/browse/HFEYP-508)

Fixing the banner notice issue caused previously delayed notices to show up immediately.  As we are only interested in notifying users who explicitly set their cookie preferences while on the cookies page, and do not care to keep those who set them otherwise informed, we will only set the message when requested which can be done using a hidden field on the form submitting the request.


guy-roberts (approved):
It works, nice

PR #353, opened 2021-07-21
Also need to check if page has parent

ebrett:
### Context

[HFEYP-489](https://dfedigital.atlassian.net/browse/HFEYP-489)

Fixes the two regressions found in the latest test run

### DIFF
--- a/app/controllers/content_controller.rb
+++ b/app/controllers/content_controller.rb
@@ -16,7 +16,7 @@ class ContentController < ApplicationController
     @markdown = GovspeakToHTML.new.translate_markdown(@page.markdown)
 
     # If in a section page and page has parent, it must be a child of existing page
-    not_found if content_section? && params["section"] != parent_slug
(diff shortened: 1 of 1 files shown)


app/controllers/content_controller.rb
```diff
@@ -16,7 +16,7 @@ def show
     @markdown = GovspeakToHTML.new.translate_markdown(@page.markdown)
 
     # If in a section page and page has parent, it must be a child of existing page
-    not_found if content_section? && params["section"] != parent_slug
+    not_found if (@page.parent || content_section?) && params["section"] != parent_slug
```
guy-roberts, line 19: Ideally we should to add tests for routing changes

guy-roberts (approved):

app/controllers/content_controller.rb
ebrett: Agreed.  Need to think about this a bit.

PR #354, opened 2021-07-22
Content Block triage

guy-roberts:
### Context
Four changes to the ContentBlock stuff

1. Removed the ContentBlock show action and view
2. Form for ContentBlock now has red line around the third field if there is an error
3. Landing page reverted to not use ContentBlock for Other useful Resources, fixed the h2 size
4. error message for missing Content field now has field name

### Guidance to review
Check that these changes work


ebrett (approved):
I might consider removing the destroy action also, if we're not allowing blocks to be deleted.

guy-roberts:
> I might consider removing the destroy action also, if we're not allowing blocks to be deleted.

Roger

ebrett (approved):

PR #357, opened 2021-07-23
Database changes for Content Blocks

ebrett:
### Context
Standalone branch for release of database changes for content blocks

I have split out the parts of main related to the database migrations and seeds. 

### Guidance to review
This is branch only for deployment to pre-prod and prod prior to the main 1.6 release (which also has these changes), so this PR will not be merged to main.  It is only for review purposes.


db/seeds/create_content_blocks.rb
```diff
@@ -0,0 +1,258 @@
...
+* change colours, contrast levels and fonts
+* zoom in up to 300% without the text spilling off the screen
+* navigate the website using just a keyboard
+* navigate the website using speech recognition software
+* listen to the website using a screen reader (including the most recent versions of JAWS and NVDA)
+
+We’ve also made the website text as simple as possible to understand.
+
+# How accessible this website is
+
```
guy-roberts, line 38: This should be a h2 

Darren noticed,  'How accessible this website is' within seed is down as h1 but should be h2.

guy-roberts (approved):

guy-roberts (approved):

db/seeds/create_content_blocks.rb
ebrett: ok, updated

dmallett1 (approved):
As prod

dmallett1 (approved):

guy-roberts (commented):
Looks good to me

I agree that we should *never* run seed data on production.

What works for me is to load it onto another version and then create the Content Block on production by pasting in the markdown manually.

This only takes a few minutes

PR #378, opened 2021-08-13
539 Card sizing alignment for mobile landscape

ebrett:
[HFEYP-539](https://dfedigital.atlassian.net/browse/HFEYP-539)

### Context
Height difference in cards for mobile landscape.

Margin bottom of zero on last child was causing issue. Need to confirm with @tunde the purpose of the 0 margin and whether it's ok to remove in all cases or if we need to target specific use cases (i.e. width where there are two columns with cards side 
by side).  The issue (different heights) is still there when it reduces to a single column but I think it is done to minimize the space between sections. If we're ok with the additional space, then this solution should work.

### Guidance to review

Before:
CleanShot 2021-08-13 at 12 32 06@2x

After:
CleanShot 2021-08-13 at 12 31 23@2x

### DIFF
--- a/app/webpacker/styles/_cards.scss
+++ b/app/webpacker/styles/_cards.scss
@@ -89,10 +89,6 @@ $card-border-hover-color: #003078;
     .eyfs-card {
       margin-bottom: govuk-spacing(3);
     }
-
-    &:last-child .eyfs-card {
-      margin-bottom: 0;
-    }
   }
 
   .eyfs-card--feature {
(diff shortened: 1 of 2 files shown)


guy-roberts (approved):
Looks like I added that in February.

I cannot remember why.

ebrett:
> Looks like I added that in February.
> 
> I cannot remember why.

Well, it does look better with less space, but the uneven heights looks worse.

PR #382, opened 2021-08-17
Try setting variant in session

ebrett:
### Context
[HFEYP-517](https://dfedigital.atlassian.net/browse/HFEYP-517)

To make the variant choice sticky, save in session and to use if present


guy-roberts (approved):
I don't know the context, but you are trying things out so probably want this approving

PR #396, opened 2021-09-09
Basic article functionality

ebrett:
### Context

[HFEYP-548](https://dfedigital.atlassian.net/browse/HFEYP-548)

Just the following functionality 

List articles, show, create, update and delete an article


app/controllers/admin/articles_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  private
+
+    def set_article
+      @article = ContentPage.find(params[:id])
+    end
+
+    def article_params
+      # category set to article in form
+      # position set to 9999 in form
```
guy-roberts, line 47: Positions have to be unique, so you'll run into validation errors

app/models/content_page.rb
```diff
@@ -12,18 +12,22 @@ class ContentPage < ApplicationRecord
   validates :title, presence: true, uniqueness: true
   validates :markdown, presence: true
 
-  validates :position, presence: true, numericality: { only_integer: true }
-  validates :position, uniqueness: { scope: :parent_id }
+  validates :position, presence: true, numericality: { only_integer: true }, unless: -> { article? }
+  validates :position, uniqueness: { scope: :parent_id }, unless: -> { article? }
```
guy-roberts, line 17: These special cases are hints that something is wrong.   Articles will still need an order in the view.  It might be overridden, but anything is better than random

app/views/admin/articles/_form.html.haml
```diff
@@ -0,0 +1,12 @@
...
+      model: [:admin, article],
+      url: article.new_record? ? admin_articles_path : admin_article_path(article),
+      local: true,
+      builder: GOVUKDesignSystemFormBuilder::FormBuilder
+    ) do |form|
+  = form.govuk_error_summary order: %i[title markdown]
+  = form.govuk_text_field :title 
+  = form.govuk_text_area :markdown
+  = form.hidden_field :category, value: 'article'
+  = form.hidden_field :position, value: 9999
```
guy-roberts, line 11: As above, I'd keep assigning positions, you don't have to use them.  Then you don't have to have the special cases when checking for position in the rest of the code

app/controllers/admin/articles_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+        redirect_to admin_article_path(@article), notice: "Article was successfully updated"
+      else
+        render :edit
+      end
+    end
+
+    def destroy
+      @article.destroy!
+      redirect_to admin_articles_path, notice: "Article was successfully destroyed"
+    end
```
guy-roberts, line 37: Could localise the notices

config/routes.rb
```diff
@@ -12,6 +12,7 @@
   namespace :admin do
     root to: "users#index"
     resources :users
+    resources :articles
   end
```
guy-roberts, line 16: They need to be under the CMS namespace, the same as the ContentPages and ContentBlocks. Otherwise these routes will show up in the public version wont they ?

I think I added this class to hide CMS routes, like the new ones to articles

  constraints CmsRouteConstraint.new do

db/migrate/20210909094348_add_category_to_content_pages.rb
```diff
@@ -0,0 +1,5 @@
+class AddCategoryToContentPages < ActiveRecord::Migration[6.1]
+  def change
+    add_column :content_pages, :category, :string
+  end
```
guy-roberts, line 4: I'd call it categories, so that a content_page can belong to more than one at a time.

config/routes.rb
ebrett: Well users is under admin and has been since the start.  Plan is to move content pages and blocks here also and then have a single app as this way is rather unconventional.

db/migrate/20210909094348_add_category_to_content_pages.rb
ebrett: Can you think of another category we'd use?

config/routes.rb
guy-roberts: They can go anywhere, but the way you've got them here does not use the CmsRouteConstraint to make sure that they will be hidden on the public site

ebrett: OK, I may need a picture drawn for me; seems like user/admin is hidden on the public site even without being in the CmsRouteConstraint.

db/migrate/20210909094348_add_category_to_content_pages.rb
ebrett: I'm not sure I want to mix together the concepts without more discussion.  For the purpose of this change each page has one category (either an article or not).  I would consider groupings of pages to be a different.

app/views/admin/articles/_article.html.haml
```diff
@@ -0,0 +1,8 @@
+%tr.govuk-table__row
+  %td.govuk-table__cell
+    = article.title
+  %td.govuk-table__cell
+    Description to come
+  %td.govuk-table__cell
```
guy-roberts, line 6: We used to have a description for ContentPage but Steve made me remove it !

db/migrate/20210909094348_add_category_to_content_pages.rb
guy-roberts: Thinking, the reason you need a category attribute, is because ContentPage inherits from Article, but you still need to filter out the ContentPages

This is a sign that something is wrong !  I don't think a ContentPage is an Article

app/models/content_page.rb
```diff
@@ -1,22 +1,13 @@
-class ContentPage < ApplicationRecord
+class ContentPage < Article 
   acts_as_tree
```
guy-roberts, line 2: I don't think a ContentPage is an Article

A clue is that you still need an attribute called category with the value 'article' to tell the two apart !

guy-roberts (approved):
I am kind of agreeing to differ on this approach

It's adding a lot of complexity that could be avoided.

app/webpacker/packs/application.js
```diff
@@ -20,6 +20,10 @@ $(document).ready(function() {
     createPreview('#markdown-editor','#markdown-render');
   });
 
+  $('#article-markdown-field').keyup(function() {
+    createPreview('#article-markdown-field', '#markdown-render');
+  });
+
```
guy-roberts, line 26: :-( 

This is where the duplication starts to take its toll

guy-roberts (approved):
Looks good to me

Its a pity that the work for drafting and publishing will have to be repeated for Articles

ebrett:
I'll make the drafting and publishing as DRY as possible - I'll get there is stages (as well as removing that ugly bit of javascript duplication)

PR #411, opened 2021-09-23
Drafts and publishing

guy-roberts:
### Context
Drafting and publishing of ContentPages

There are several tickets in Jira

531, 533, 535, 438, 550, 569, 570, 571, 572

Several of these tickets have out of date Figma pictures. Changes were agreed and the tickets not updated. This was mostly about moving buttons on forms.  

Note the difference between the status, Published/Unpublished and the existence of draft versions

These are two different ideas

ContentPages are now created as unpublished
When you edit a page, you'll first be take to a table of versions
From here you can edit a draft, in which case changed are saved to that draft
Or you can edit a published page, in which case changes are saved to a new draft
Drafts can be deleted
Deleting the last draft of an unpublished page will remove the page
There is no way to un-publish a page at the moment

### Guidance to review
You can try this out here

https://eyfs-cms-review-pr-411.london.cloudapps.digital/cms/pages
guy@education.gov.uk / GuyGuy2021!!

Go through the life cycle of a page and check that it matches the Jira specs (though note that some are out of date)

After publishing a page, you'll need to do a hard refresh to tell the browser to refresh its cached version of the page


ebrett (approved):
Looks good to me!

PR #468, opened 2021-10-31
Fix article images

tunde-anddigital:
## Ticket and context

Ticket: [HFEYP-547](https://dfedigital.atlassian.net/browse/HFEYP-547)
## Tech review
Updates to ticket 547 - Fixed featured image size for the article page (spans full width and is responsive)


db/schema.rb
```diff
@@ -124,6 +124,7 @@
     t.integer "previous_id"
     t.boolean "is_published", default: false
     t.string "author"
+    t.boolean "navigation", default: true
```
ebrett, line 127: I don't think this should be here - is this not from a previous attempt at doing article pages?

ebrett (changes requested):
remove the schema change

ebrett (approved):

PR #471, opened 2021-11-02
Make links format correctly

ebrett:
## Ticket and context

Ticket: [HFEYP-596](https://dfedigital.atlassian.net/browse/HFEYP-596)

## Tech review

Some of the links were missing the govuk-link class - I've added them to the links in question

### DIFF
--- a/app/views/content_pages/_details_for_a_single_page.erb
+++ b/app/views/content_pages/_details_for_a_single_page.erb
@@ -6,7 +6,7 @@
       <span class="links">
         <%= link_to 'Versions', versions_content_page_path(page_to_render), { class: "govuk-link" } %>
         <% if !page_to_render.parent_id %>
-          |<%= link_to 'Add a child page', new_content_page_path(:parent_id => page_to_render.id), { class: "govuk-link" } %>
+          |<%= link_to 'Add a child page', new_content_page_path(parent_id: page_to_render.id), { class: "govuk-link" } %>
         <% end %>
         <% if (page_to_render.is_published) %>
           | <%= link_to 'Preview', path_for_this_page(page_to_render), { class: "govuk-link", target: "_blank" } %>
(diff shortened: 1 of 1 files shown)


app/views/content_pages/_details_for_a_single_page.erb
```diff
@@ -4,12 +4,12 @@
     <p class="govuk-heading-m">
       <%= page_to_render.title %>
       <span class="links">
-        <%= link_to 'Versions', versions_content_page_path(page_to_render), { :class => "govuk-link" } %>
+        <%= link_to 'Versions', versions_content_page_path(page_to_render), { class: "govuk-link" } %>
         <% if !page_to_render.parent_id %>
-          |<%= link_to 'Add a child page', new_content_page_path(:parent_id => page_to_render.id) %>
+          |<%= link_to 'Add a child page', new_content_page_path(:parent_id => page_to_render.id), { class: "govuk-link" } %>
```
reggieb, line 9: Very minor - which you can ignore - but is it work updating the `parent_id` assignment:
```ruby
|<%= link_to 'Add a child page', new_content_page_path(parent_id: page_to_render.id), { class: "govuk-link" } %>
```

reggieb (approved):
Looks good to me.

Perhaps as a tech debt it would be good to have a helper method like this:
```ruby
def govuk_link_to(name = nil, options = nil, html_options = {}, &block)
  html_options.merge(class: "govuk-link")
  link_to(name = nil, options = nil, html_options = nil, &block)
end
```
We could then do:
```ruby
govuk_link_to 'Versions', versions_content_page_path(page_to_render)
```

ebrett:
> Looks good to me.
> 
> Perhaps as a tech debt it would be good to have a helper method like this:
> 
> ```ruby
> def govuk_link_to(name = nil, options = nil, html_options = {}, &block)
>   html_options.merge(class: "govuk-link")
>   link_to(name = nil, options = nil, html_options = nil, &block)
> end
> ```
> 
> We could then do:
> 
> ```ruby
> govuk_link_to 'Versions', versions_content_page_path(page_to_render)
> ```

I considered that, but I think there are a few cleanups like this we could do, and I'd like to create a ticket for them and do it all at once.

PR #473, opened 2021-11-03
Hfeyp 606 cms author date

tunde-anddigital:
## Ticket and context

Ticket: [HFEYP-606](https://dfedigital.atlassian.net/browse/HFEYP-606)
## Tech review
- Add unpublished_at date to tab in article cms
- Add created_at date to drafts tab in article cms
- Conditionally show the correct date depending on article state (published/unpublished/draft)


db/schema.rb
```diff
@@ -124,6 +125,7 @@
     t.integer "previous_id"
     t.boolean "is_published", default: false
     t.string "author"
+    t.boolean "navigation", default: true
```
reggieb, line 128: This field looks like it has slipped in from another branch. This line should be removed from the schema.

app/controllers/admin/articles_controller.rb
```diff
@@ -15,6 +15,7 @@ def new
     def create
       @article = Article.new(article_params)
       if @article.save
+        @article.touch(:created_at)
```
reggieb, line 18: Using `touch` here will cause a second call to the database. Also if that second call fails the redirect will continue without the `created_at` being updated.

I'd suggest updating line 16 to:
```ruby
@article = Article.new(article_params.merge(created_at: Time.now.utc))
```
Rather than adding the `touch` call. That way, there will only be one write call to the database, and the redirect will only occur if that is successful.

app/controllers/admin/articles_controller.rb
```diff
@@ -44,6 +45,7 @@ def publish
 
     def unpublish
       @article.unpublished!
+      @article.touch(:unpublished_at)
```
reggieb, line 48: I can't see where the `Article` method `unpublished!` is defined, but that looks like a write operation. Could that be updated to include an update of the `unpublished_at` field?

app/views/admin/articles/_article.html.haml
```diff
@@ -2,7 +2,12 @@
   %td.govuk-table__cell
     = article.description
   %td.govuk-table__cell
-    = article.published_at&.strftime("%d-%m-%Y")
+    - if article.published?
+      = article.published_at&.strftime("%d-%m-%Y")
+    - elsif article.unpublished?
+      = article.unpublished_at&.strftime("%d-%m-%Y")
+    - else
+      = article.created_at&.strftime("%d-%m-%Y")
```
reggieb, line 10: If you add an initializer: `config/initializers/time_formats.rb` with this content:
```ruby
Time::DATE_FORMATS[:date_only] = "%d-%m-%Y"  # 04-11-2021
```
Here you could do:
```ruby
    - if article.published?
      = article.published_at&.to_s(:date_only)
    - elsif article.unpublished?
      = article.unpublished_at&.to_s(:date_only)
    - else
      = article.created_at&.to_s(:date_only)
```
Which will aid consistency and better communicate intent.

app/controllers/admin/articles_controller.rb
tunde-anddigital: Not sure myself where `unpublished?` is defined. There's a file `article_policy` in `app/policies` which I think might need to be modified, but again I'm not too sure. A lot of this is Brett's code so I was just working on some additions in this pr

reggieb: OK - leave it as is then.

reggieb (approved):
LGTM

app/controllers/admin/articles_controller.rb
ebrett: `enum` in rails defines a bunch of helper methods (i.e. `unpublished!` `unpublished?` and scope methods like `Article.unpublished`. I thought this was common knowledge, but Guy was also unaware - perhaps should avoid using in the future to avoid these type of surprises?  I just find them really convenient.

And yes, you can override it to combine the database writes, but I was going for clarity here (and following guy's lead of putting everything in the controller rather than the model) - happy for it to move to the model and for the method to be overwritten.

ebrett (approved):
Happy for you to make the suggested changes (also to the published action) or we can save for a later update.

app/controllers/admin/articles_controller.rb
reggieb: Yes - I see it now. Thanks @ebrett 

I think that setting `unpublished_at` is so tightly coupled to `unpublished!` that the two should be tied together in the model. Perhaps like this in the model:

```ruby
def unpublished!
 self.unpublished_at = Time.now.utc
 super
end
```

PR #480, opened 2021-11-09
Markdown is already sanitized

ebrett:
## Ticket and context

 No ticket - fixing an issue with auto-generated ID's being removed from heading

* the second sanitize was removing the id from the headings


reggieb (approved):
LGTM

PR #493, opened 2021-11-18
HFEYP-580 article order

tunde-anddigital:
## Ticket and context

Ticket: [HFEYP-580](https://dfedigital.atlassian.net/browse/HFEYP-580)
## Tech review
- Allow users to sort Articles created in the CMS by their  `description` or by the `date`
- Default order for articles in CMS is alphabetical ascending
- Users can click headings to change the order articles are displayed

### Code quality checks

1. Go to https://eyfs-cms-review-pr-493.london.cloudapps.digital/cms/pages
2. Log in (Go to Jira ticket 580 for credentials)
3. Click on the "Articles" link in the navigation bar
4. Click on the table headings (Description / Published / Created / Unpublished) to sort articles by the selected column (in ascending order)


app/views/admin/articles/index.html.haml
```diff
@@ -8,9 +8,9 @@
       %thead.govuk-table__head
         %tr.govuk-table__row
           %th.govuk-table__header{scope: 'col', class: 'govuk-!-width-one-quarter'}
-            Description 
+            = link_to("Description", anchor: "published", :sort => "description")
```
reggieb, line 11: A nicer syntax for hashes where the keys are symbols is `key: value` rather than `:key => value` so I'd suggest all the `sort` calls in the `link_to` elements in this view be defined in this way. So:
```ruby
= link_to("Description", anchor: "published", sort: "description")
```

reggieb (changes requested):
I think the current code change needs a very minor tweak.

However, what missing are tests of this new behaviour. I'd suggest adding to `spec/features/admin/admin_articles_spec.rb` or creating a new file: `/home/rob/web/dfe/early-years-foundation-reform/spec/features/admin/articles/sorting_spec.rb` (the more I think about it, the more I like the latter option)

reggieb (approved):
Excellent

PR #495, opened 2021-11-22
Hfeyp 620 increase spec test coverage above 95.94%

reggieb:
## Ticket and context

Ticket: [HFEYP-620](https://dfedigital.atlassian.net/browse/HFEYP-620)
### Is there anything that the code reviewer should know?
This mostly consists of adding spec where they are missing, but also removing unused code.

### Code quality checks

This change should not change the behaviour of the review app


reggieb:
The current version takes the code coverage to 99.96%. Just the [GovspeakDecorator::HtmlSanitizerDecorator#sanitize method](https://github.com/DFE-Digital/early-years-foundation-reform/blob/develop/app/models/govspeak_decorator.rb#L76) is not covered. I haven't been able to work out meaning for spec that would use the code in that line.

tunde-anddigital (commented):
Looks all good to me

tunde-anddigital (approved):

PR #501, opened 2021-12-01
Hide support articles section unless there are published articles

reggieb:
## Ticket and context

Ticket: [HFEYP-621](https://dfedigital.atlassian.net/browse/HFEYP-621)

Note that the requirement has change. It is now to hide the section if there are no published articles, rather than using an environment configuration option.

### Code quality checks

2. The articles section should not be present (unless someone has published an article)
3. Create an article in CMS and publish it
4. Return to the home page and the articles section should be present.
5. Return to CMS and unpublish all the articles there
6. Return to the home page and the articles section should be absent.

### DIFF
--- a/app/views/content/index.html.erb
+++ b/app/views/content/index.html.erb
@@ -32,19 +32,21 @@
         <% end %>
       </ul>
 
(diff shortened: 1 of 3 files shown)


tunde-anddigital (approved):
Looks all good, tested in the review app also and works as required

PR #507, opened 2021-12-03
Remove merging from deploy to production github workflow

reggieb:
Also update sandbox workflow to closer to production, so it can act as test bed

This is a maintenance update. With this change the workflow will change so that `develop` is merged into `main` outside of the deployment script (via a pull request). Once that is done, the deploy to production workflow can be triggered via actions -> 'Deploy to Production'


tunde-anddigital (approved):

reggieb:
This PR has been superseded by this https://github.com/DFE-Digital/early-years-foundation-reform/pull/517
So closing

PR #509, opened 2021-12-06
Merging develop into main for deploying v1.9.5

reggieb:
Merging main branches


tunde-anddigital (approved):

reggieb:
On merging this PR I chose the rebase option which was a mistake. It created a separate branch away from develop rather than merging develop with main so that they were at the same point.

To correct the issue, I pull the change locally, hard reset back to the original divergence between main and develop, merged locally, turned off forced push protection for the main branch, force pushed main, and then turned back on the forced push protection for main.

PR #534, opened 2022-01-06
HFEYP-610 Enable User feedback

reggieb:
## Ticket and context

Ticket: [HFEYP-610](https://dfedigital.atlassian.net/browse/HFEYP-610)

### Feedback panel

- Users can view the Feedback panel across all website page footer as per design from ife.akinbolaji 
- ‘Give feedback’ button should take user to the Google form[Help for Early Years Providers ]
- User click on ‘Yes’ button should display the message “Thank you for your feedback”
- User click on ‘No’ should expand the form with options to:
  - view the text as per wireframe
  - “Close” button to close the form and return to original state, and retain focus on the ‘No’ button as per screenshot
  - and “Give feedback” to take user to the Google form [Help for Early Years Providers ] in a new tab, and retain focus on the “Give feedback” button

## Tech review
There is a little tiding up of the spacing in layouts included.

### Code quality checks

2. The feedback form should be visible at the bottom of the page. Test the functionality described above.
3. Navigate to another page and make sure the functionality works there too.


ebrett (commented):
I wonder if it might be better (assuming we go ahead with going to a single app, so the public app can write to the database) to go with the gem https://github.com/ankane/ahoy for tracking user events instead, so as to avoid asking repeatedly their opinion for the same page?

ebrett (commented):
Also, the style of the feedback block between content pages and articles is very different - but that may be solved by @agoldstone93 fix for Govspeak.

reggieb:
@ebrett I'll leave this as is, as it's not going to work with the current production setup - specifically no database write access for main app.

ebrett:
@reggieb Can we go ahead a close this PR?  Whatever it turns out to be, it's going to be very different from this solution...

reggieb:
@ebrett I'm happy to close this. It will be easy enough to reopen if needs be.

PR #538, opened 2022-01-07
HFEYP-626 Adds alt text to article feature image with helper

ebrett:
Ticket: [HFEYP-626](https://dfedigital.atlassian.net/browse/HFEYP-626)

Look at feature images on articles in the admin area preview page and on the public portion of the site.
Thumbnail images on the articles page also now have alt text included.  

In addition, the title of the page now shows as the title of the images.

### Code quality checks

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,7 +15,7 @@ module ApplicationHelper
     return unless page.featured_image.attached?
 
     tag.div(class: "page-img-container") do
-      image_tag(url_for(page.featured_image), class: "page-responsive-img", alt: page.featured_alt_text, title: page.title, width: "500px")
+      image_tag(url_for(page.featured_image), class: "page-responsive-img", alt: page.featured_alt_text, title: page.featured_alt_text, width: "500px")
     end
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -12,10 +12,10 @@ module ApplicationHelper
   end
 
   def feature_image(page:)
-    if page.featured_image.attached?
-      tag.div(class: "page-img-container") do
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -10,4 +10,12 @@ def nav_link_to(title:, path:)
...
     end
   end
+
+  def feature_image(page:)
+    if page.featured_image.attached?
+      tag.div(class: "page-img-container") do
+        image_tag(url_for(page.featured_image), class: "page-responsive-img", alt: page.featured_alt_text, title: page.title, width: "500px")
+      end
+    end
+  end
```
reggieb, line 20: I'd suggest:
```ruby
def feature_image(page:)
  return unless page.featured_image.attached?

  tag.div(class: "page-img-container") do
    image_tag(url_for(page.featured_image), class: "page-responsive-img", alt: page.featured_alt_text, title: page.title, width: "500px")
  end
end
```
Also - should the width be defined in CSS via `.page-responsive-img` rather than hard coded into the HTML.

ebrett: I'll add the guard - the width in the html is for the browser and not for the css (see https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/)

reggieb: Re. My comment about width: My understanding of best practice is that it is better to define an element's appearance (including width) via CSS rather than directly in HTML. So:

css: `.page-responsive-img { width: 500px }`
html: `...</img>`

Rather than:
html: `...</img>`

That makes it easier to do things like change the width depending on the surrounding or window size.

reggieb (approved):
I'm happy for my comments on width to be ignored. Change looks good to me.

PR #550, opened 2022-01-13
HFEYP-604 Update styles for contents list

ebrett:
Copy and pasted from govuk_publishing_component as was
unable to use from gem as it does not work with webpacker

Ticket: [HFEYP-604](https://dfedigital.atlassian.net/browse/HFEYP-604)

### Is there anything that the code reviewer should know?

I did try to add the styles from the above gem (which is a dependency of the govspeak gem we use)

However, it appears to not work with webpacker.  Suggest we upgrade to Rails 7 and see if we can
remove the need for webpacker and start using the styles in govspeak and in the above gem, rather
than the copy and pasted styles from those gems.  But this seems like a bigger refactoring then
we want to pursue right now.

https://github.com/alphagov/govuk_publishing_components/issues/505

### Code quality checks

## Product review

To see the styles, enter the following in the markdown of a content page

```
{:.contents-list__list}
* {:.contents-list__list-item .contents-list__list-item--dashed} [Heading](#id-for-anchor)
```


app/webpacker/styles/_contents-list.scss
```diff
@@ -0,0 +1,98 @@
...
+  // orange border around the em dash.
+  // Use inline-block and vertical alignment to
+  // fix focus styles
+  //
+  // https://github.com/alphagov/government-frontend/pull/420#issuecomment-320632386
+  .lte-ie8 & .gem-c-contents-list__link {
+    display: inline-block;
+    vertical-align: top;
+  }
+}
```
reggieb, line 98: Need an empty line here

reggieb (approved):
Looks fine to me

PR #586, opened 2022-02-01
Hfeyp 653 asset folder components

agoldstone93:
## Ticket and context

Ticket: [HFEYP-653](https://dfedigital.atlassian.net/browse/HFEYP-653)
## Tech review

- Added a column to the content asset model called 'page_name'
- Added radio buttons to form to edit asset page to set correct page name
- Added accordion to asset page, which displays a section for each content page title
- Added a copy link button to each image

### Code quality checks

### Is there anything that the product reviewer should know?
The design hasn't been followed exactly for a few reasons:
- the prototype is quite old
- the govuk accordion design has changed
- some radio buttons for selection needed to be added to the edit page to make it make sense

1. Click [the link to review app](https://eyfs-cms-review-pr-586.london.cloudapps.digital/admin/assets)
2. Add an asset. Select which accordion section it should be in by selecting the 'page title'.
3. Verify that it is in the correct section.
4. Edit the page title and verify that it moves to the correct section

### DIFF
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -4,6 +4,7 @@ module Admin
 
     def index
       @content_assets = ContentAsset.all
+      @content_pages = ContentPage.top_level.order_by_position
     end
 
     def show; end
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -1,7 +1,7 @@
 module Admin
   class ContentAssetsController < AdminController
     before_action :set_content_asset, only: %i[show edit update destroy]
-    before_action :set_folder_options, only: %i[index new edit]
+    before_action :set_folder_options
 
     def index
       @content_assets = ContentAsset.all
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -60,7 +60,7 @@ module Admin
 
     # Only allow a list of trusted parameters through.
     def content_asset_params
-      params.require(:content_asset).permit(:title, :asset_file, :alt_text, :page_name)
+      params.require(:content_asset).permit(:title, :asset_file, :alt_text, :content_page_id)
     end
   end
 end
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -4,7 +4,6 @@ module Admin
 
     def index
       @content_assets = ContentAsset.all
-      @content_pages = ContentPage.top_level.order_by_position
     end
 
     def show; end
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -1,7 +1,7 @@
 module Admin
   class ContentAssetsController < AdminController
     before_action :set_content_asset, only: %i[show edit update destroy]
-    before_action :set_folder_options
+    before_action :set_folder_options, only: %i[index edit new create update]
 
     def index
       @content_assets = ContentAsset.all
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -1,6 +1,7 @@
 module Admin
   class ContentAssetsController < AdminController
     before_action :set_content_asset, only: %i[show edit update destroy]
+    before_action :set_folder_options, only: %i[index new edit]
 
     def index
       @content_assets = ContentAsset.all
@@ -61,6 +62,10 @@ module Admin
     # Only allow a list of trusted parameters through.
     def content_asset_params
       params.require(:content_asset).permit(:title, :asset_file, :alt_text, :content_page_id)
+    end
+
+    def set_folder_options
+      @folder_options = ContentPage.top_level.order_by_position.all + [OpenStruct.new(id: 0, title: "Other")]
     end
   end
 end
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -61,7 +61,7 @@ module Admin
 
     # Only allow a list of trusted parameters through.
     def content_asset_params
-      params.require(:content_asset).permit(:title, :asset_file, :alt_text)
+      params.require(:content_asset).permit(:title, :asset_file, :alt_text, :page_name)
     end
   end
 end
--- a/app/controllers/admin/content_assets_controller.rb
+++ b/app/controllers/admin/content_assets_controller.rb
@@ -65,7 +65,7 @@ module Admin
     end
 
     def set_folder_options
-      @folder_options = ContentPage.top_level.order_by_position.all + [OpenStruct.new(id: 0, title: "Other")]
+      @folder_options = ContentPage.top_level.order_by_position.all + [OpenStruct.new(id: nil, title: "Other")]
     end
   end
 end
--- a/app/helpers/content_pages_helper.rb
+++ b/app/helpers/content_pages_helper.rb
@@ -1,5 +1,0 @@
-module ContentPagesHelper
-  def get_all_content_pages
-    ContentPage.top_level.order_by_position
-  end
-end
--- a/app/helpers/content_pages_helper.rb
+++ b/app/helpers/content_pages_helper.rb
@@ -1,5 +1,5 @@
 module ContentPagesHelper
   def get_all_content_pages
-    @content_pages = ContentPage.top_level.order_by_position
+    ContentPage.top_level.order_by_position
   end
 end
--- a/app/helpers/content_pages_helper.rb
+++ b/app/helpers/content_pages_helper.rb
@@ -1,0 +1,5 @@
+module ContentPagesHelper
+  def get_all_content_pages
+    @content_pages = ContentPage.top_level.order_by_position
+  end
+end
--- a/app/views/admin/content_assets/index.html.erb
+++ b/app/views/admin/content_assets/index.html.erb
@@ -15,8 +15,9 @@
         </thead>
 
         <tbody class="govuk-table__body">
-          <% @content_assets.each do |content_asset| %>
-            <% if content_asset.content_page_id == content_page.position %>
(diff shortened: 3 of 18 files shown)


app/helpers/content_pages_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ContentPagesHelper
+  def get_all_content_pages
```
reggieb, line 2: Using a description of the action used to create an object to name it, isn't a very Rails way of doing things. It's better practice to have object names describe the object they returns. So this would be a better name:
```ruby
def top_level_content_pages_by_position
```
However, I'd also question why this is being done in a helper method, which should really only be used to decorate/modify view behaviour. I'd suggest that a better option would be to create a new scope in the content page model:
```ruby
scope :top_level_by_position, -> { top_level.order_by_position }
```
Then in the views, use `ContentPage.top_level_by_position` where it is needed.

app/views/admin/content_assets/index.html.erb
```diff
@@ -1,25 +1,36 @@
...
-      </tr>
+        <tbody class="govuk-table__body">
+          <% @content_assets.each do |content_asset| %>
+            <% if content_asset.page_name.to_i == content_page.position %>
+              <tr class="govuk-table__row">
+                <td class="govuk-table__cell"><%= image_tag(url_for(content_asset.asset_file), class: 'govuk-!-width-full') %></td>
+                <td class="govuk-table__cell" ><%= content_asset.title %></td>
+                <td class="govuk-table__cell">
+                  <%= govuk_button_link_to 'Edit', edit_admin_content_asset_path(content_asset), secondary: true %>
+                  <%= button_tag "Copy to clipboard", type: 'button', class: "govuk-button govuk-button--secondary", onClick: "navigator.clipboard.writeText('#{polymorphic_url(content_asset.asset_file)}')" %>
```
reggieb, line 25: Pedantic, but lowercase on click key works fine and matches rails contentions (to use snake case rather than camel). Also I'd suggest splitting this over a few lines to make it clearer:
```ruby
 <%= button_tag(
       "Copy to clipboard", 
       type: 'button', 
       class: "govuk-button govuk-button--secondary", 
       onclick: "navigator.clipboard.writeText('#{polymorphic_url(content_asset.asset_file)}')"
     ) %>
```

reggieb (commented):
A couple of very small tweaks suggested

app/helpers/content_pages_helper.rb
ebrett: I'd also just chain it, that's also perfectly acceptable `ContentPage.top_level.order_by_position`

db/schema.rb
```diff
@@ -92,6 +92,7 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.string "alt_text", default: "", null: false
+    t.string "page_name"
```
ebrett, line 95: Rather than linking an asset to a ContentPage by it's title, can add a `content_page_id` reference - then have `has_one :content_page` (or many if we decide to go that way, in which case a join table would be needed)

ebrett (changes requested):

app/helpers/content_pages_helper.rb
agoldstone93: i've made this suggested change

app/views/admin/content_assets/index.html.erb
agoldstone93: i've made this change

db/schema.rb
agoldstone93: i've made the change that i think you've suggested, but it might still not be quite right. Let me know if I should do some more work on it

reggieb: Ummm.... shouldn't it be a `belongs_to` on the `ContentAsset` model. Then a `has_one` or `has_many` on the `ContentPage` model.

spec/factories/content_assets.rb
```diff
@@ -2,6 +2,7 @@
   factory :content_asset do
     title { Faker::Lorem.sentence(word_count: 4).to_s }
     alt_text { Faker::Lorem.sentence(word_count: 4).to_s }
+    content_page_id { rand(1..100) }
```
reggieb, line 5: No - don't do that - this will cause unexpected behaviour (when the id matches an existing content page). If the content page relationship is set up correctly, this should work:
```ruby
factory :content_asset do
  content_page
```
setting the association name here will create a new associated object and attach to the new model instance.

spec/models/content_asset_spec.rb
```diff
@@ -7,6 +7,7 @@
     before(:each) do
       content_asset.title = "Title"
       content_asset.alt_text = "Sample Alt Text"
+      content_asset.content_page_id = 1
```
reggieb, line 10: Again - don't create lone ids like this. Instead:
```ruby
content_asset.content_page = create(:content_page)
```

spec/views/admin/content_assets/index.html.erb_spec.rb
```diff
@@ -3,8 +3,8 @@
 RSpec.describe "admin/content_assets/index", type: :view do
   before(:each) do
     @content_assets = [
-      FactoryBot.create(:content_asset, title: "title 1", alt_text: "hello"),
-      FactoryBot.create(:content_asset, title: "title 2", alt_text: "hello 2 u"),
+      FactoryBot.create(:content_asset, title: "title 1", alt_text: "hello", content_page_id: 1),
+      FactoryBot.create(:content_asset, title: "title 2", alt_text: "hello 2 u", content_page_id: 2),
```
reggieb, line 7: If you are not using the content page anywhere in the spec, just let the factory do its job (that is, build an associated object unless directed to do otherwise). So just leave it as:
```ruby
  FactoryBot.create(:content_asset, title: "title 1", alt_text: "hello"),
  FactoryBot.create(:content_asset, title: "title 2", alt_text: "hello 2 u"),
```
The factory will do the rest for you.

db/schema.rb
ebrett: Yes, sorry, follow Rob's lead - I got it backwards.  ContentAsset belongs to the ContentPage and ContentPage has_many many ContentAssets.

spec/views/admin/content_assets/new.html.erb_spec.rb
```diff
@@ -3,6 +3,10 @@
 RSpec.describe "admin/content_assets/new", type: :view do
   before(:each) do
     assign(:content_asset, FactoryBot.build(:content_asset))
+    @folder_options = [
```
ebrett, line 6: Just for reference, you can also do `FactoryBot.create_list(:content_page, 2)`

ebrett (approved):

app/views/admin/content_assets/index.html.erb
```diff
@@ -1,25 +1,34 @@
...
+            <th scope="col" class="govuk-table__header govuk-!-width-one-quarter">Image</th>
+            <th scope="col" class="govuk-table__header govuk-!-width-one-half">Name</th>
+            <th scope="col" class="govuk-table__header govuk-!-width-one-quarter">Action</th>
+          </tr>
+        </thead>
+        <tbody class="govuk-table__body">
+          <% content_assets = content_page.content_assets %>
+          <% if content_assets %>
+            <% content_assets.each do |content_asset|%>
+              <%= render partial: 'table_body', locals: { content_asset: content_asset } %>
```
reggieb, line 19: FYI. If you are rending a partial, as it is the default option, [there is a simpler syntax](https://api.rubyonrails.org/classes/ActionView/PartialRenderer.html#class-ActionView::PartialRenderer-label-Rendering+the+default+case). This is the equivalent:
```ruby
<%= render 'table_body',  content_asset: content_asset %>
```

app/views/admin/content_assets/index.html.erb
```diff
@@ -1,25 +1,34 @@
...
+            <% end %>
+          <% else %>
+            <% @content_assets.each do |content_asset| %>
+              <% if content_asset.content_page == nil %>
+                <%= render partial: 'table_body', locals: { content_asset: content_asset } %>
+              <% end %>
+            <% end %>
+          <% end %>
```
reggieb, line 27: Ummm .... looking at the logic in this `if` statement, I think by using a couple of ORs, the code can be simplified and the duplication removed:
```ruby
<% content_assets = content_page.content_assets || @content_assets %>
<% @content_assets.each do |content_asset| %>
  <% if content_page.content_assets || content_asset.content_page.nil? %>
    <%= render partial: 'table_body', locals: { content_asset: content_asset } %>
  <% end %>
<% end %>
```

spec/views/admin/content_assets/new.html.erb_spec.rb
reggieb: As this code is being loaded via rails_helper (and `spec/support/factory_bot.rb`)
```ruby
RSpec.configure do |config|
  config.include FactoryBot::Syntax::Methods
end
```
You don't need to specify `FactoryBot` to use its methods. So this will work too:
```ruby
@folder_options = create_list(:content_page, 2)
```

app/views/admin/content_assets/index.html.erb
agoldstone93: This suggestion doesn't quite work. This changes the logic - now all the assets are displayed in each folder rather than for their specific folder (apart from for 'other' - these display as expected)

agoldstone93: Yeah I know. That's the change I made, but now I'm getting all of the sections, except "Other", showing all of the images rather than just the images in that particular section

reggieb: OK - leave it as is!

app/controllers/admin/content_assets_controller.rb
```diff
@@ -1,6 +1,7 @@
 module Admin
   class ContentAssetsController < AdminController
     before_action :set_content_asset, only: %i[show edit update destroy]
+    before_action :set_folder_options
```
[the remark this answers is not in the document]
agoldstone93: I've updated this to be available to index, edit, new, create and  update

reggieb (approved):

PR #597, opened 2022-02-10
Hfeyp 221 navigation first page and last pages dont loop

agoldstone93:
## Ticket and context

Ticket: [HFEYP-221](https://dfedigital.atlassian.net/browse/HFEYP-221)
## Tech review

Made changes to content page model so first and last pages don't connect to each other. The first content page should not have a previous page and only a 'next page' link should be visible. The final content page should not have a next page and only a 'previous page' link should be visible.

The change will only come into effect when a page is created or a position attribute changes (or by manually calling ```ContentPage#reorder``` method)

### Code quality checks

1. Go to [first content page in review app](https://eyfs-review-pr-597.london.cloudapps.digital/get-help-to-improve-your-practice)
2. Verify that there is no "previous page" link visible
3. Go to [last content page in review app ](https://eyfs-review-pr-597.london.cloudapps.digital/communication-and-language/listening-and-understanding)
4. Verify that there is no "next page" link visible

### DIFF
--- a/app/models/content_page.rb
+++ b/app/models/content_page.rb
@@ -55,11 +55,11 @@ class ContentPage < ApplicationRecord
   end
 
   def next_page
-    if next_id then ContentPage.find next_id end
+    ContentPage.find(next_id) if next_id
   end
 
   def previous_page
-    if previous_id then ContentPage.find previous_id end
+    ContentPage.find(previous_id) if previous_id
   end
 
   def navigation
--- a/app/models/content_page.rb
+++ b/app/models/content_page.rb
@@ -55,11 +55,11 @@ class ContentPage < ApplicationRecord
   end
 
   def next_page
-    ContentPage.find next_id
+    if next_id then ContentPage.find next_id end
   end
 
   def previous_page
-    ContentPage.find previous_id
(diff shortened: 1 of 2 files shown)


app/models/content_page.rb
```diff
@@ -55,11 +55,11 @@ def set_slug_from_title
   end
 
   def next_page
-    ContentPage.find next_id
+    if next_id then ContentPage.find next_id end
```
reggieb, line 58: I more ruby way to write this is:
```ruby
ContentPage.find(next_id) if next_id
```

Or even:
```ruby
return unless next_id

ContentPage.find(next_id)
```

app/models/content_page.rb
```diff
@@ -55,11 +55,11 @@ def set_slug_from_title
...
   end
 
   def previous_page
-    ContentPage.find previous_id
+    if previous_id then ContentPage.find previous_id end
```
reggieb, line 62: Again:
```ruby
ContentPage.find(previous_id) if previous_id
```

reggieb (commented):
A small modification to the way a conditional is written and this will be fine.

PR #653, opened 2023-10-03
ER 782 Create app service to host web application dev subscription

ebrett:
## Ticket and context
[ER-782] Get the webapp up

### DIFF
--- a/.github/workflows/azure-deploy-dev.yml
+++ b/.github/workflows/azure-deploy-dev.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - ER-*
+      - main
     paths-ignore:
       - '**/*.md'
       - .docker*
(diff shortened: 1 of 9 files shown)


sunny-sidhu-and (approved):

.github/workflows/azure-deploy-dev.yml
```diff
@@ -0,0 +1,125 @@
+name: 'HfEYP App Deploy [Azure - DEV]'
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths-ignore:
```
peterdavidhamilton, line 8: The array of files to ignore might want looking at. Some of these files don't exist in this project and should be removed but others might also need to be added in their place.

sunny-sidhu-and: Good catch. Agree on this point.

ebrett: I did look at them and removed the ones below - the ones I kept were either there or could be added but should probably still be ignored.

      - .pa11yci
      - .yardopts
      - uml/*

PR #658, opened 2023-10-12
Add Dockerfile.azure with SSH, remove CONTAINER_ID and cms constraint

ebrett:
[ER-781]

- CONTAINER_ID was only used for Google Optimize, which is discontinued
- SSH will be needed for ongoing maintenance by developers
- Remove `cms` constraint on application

see https://learn.microsoft.com/en-gb/azure/app-service/configure-custom-container?tabs=alpine&pivots=container-linux#enable-ssh


ebrett:
@peterdavidhamilton @jack-coggin @sunny-sidhu-and I am able to SSH onto the machine now with this PR. However, the review app creation is broken. Holding off on merging this for now, but let's discuss on Monday.  In the meantime, please take a look and let me know if you think it looks alright.

sshd_config
```diff
@@ -0,0 +1,12 @@
+Port 			2222
```
peterdavidhamilton, line 1: Is this from https://learn.microsoft.com/en-us/azure/app-service/configure-custom-container?tabs=debian&pivots=container-linux#enable-ssh ?

@sunny-sidhu-and is this why https://eyrecovery-dev.scm.azurewebsites.net/webssh/host doesn't appear to work?

ebrett: https://hfeyp-dev.scm.azurewebsites.net/webssh/host

sunny-sidhu-and (approved):

PR #712, opened 2024-02-23
Use page for home cards

ebrett:
- Make side navigation work
- move cms/migrate
- add pages logic to content upload
- add resources for call to actions
- breadcrumbs


nomasprime (approved):

nomasprime (approved):

app/helpers/content_helper.rb
```diff
@@ -56,4 +56,70 @@ def govuk_input_classes(input_type, errors: nil)
     classes << "govuk-input--error" if errors.present?
     classes
   end
+
+  #    ul.dfe-vertical-nav__section
```
nomasprime, line 60: Can we remove?

nomasprime (changes requested):

PR #771, opened 2024-04-11
ER-920 Mobile side menu

martikat:
[ER-920](https://dfedigital.atlassian.net/browse/ER-920)

This also includes the fix for [ER-1009](https://dfedigital.atlassian.net.mcas.ms/browse/ER-1009)


peterdavidhamilton:
@martikat 

- [ ] entire stylesheets from the old design are still present and should be removed
- [ ] the active links aren't needed and the burger menu is always visible so the use of a conditional is not appropriate for the new design
- [ ] if there is a JS issue for the accordion action on the nav links let's confirm that on a prod deployment

app/models/page.rb
```diff
@@ -45,7 +45,7 @@ def breadcrumbs
     list = [self, self&.parent, self&.parent&.parent, self&.parent&.parent&.parent].compact.reverse
     list.shift
     list.each_with_object({ 'Home' => '/' }) do |obj, memo|
-      memo[obj.title] = obj.path
```
peterdavidhamilton, line 48: This code looks like `#footer_path`, is there a DRY-er way of doing this?

app/components/header_component.html.slim
```diff
@@ -7,13 +7,19 @@
...
       = link_to service_name, service_url, class: 'dfe-header__link dfe-header__link--service'
     #content-header.dfe-header__content.dfe-header-f-dtonly.dfe-header-f-toplinks
-      .dfe-header-f-top-links
-        / ul.dfe-header__action-links
-        /   - action_links.each do |item|
-        /     = item
       .dfe-header__menu
         button#toggle-menu.dfe-header__menu-toggle aria-controls='header-navigation' aria-expanded='false' data-action='click->reveal#toggle'
           | Menu
+    / This logic has been commented out as it is not required currently
```
peterdavidhamilton, line 13: Could you check that this way of commenting doesn't result in visible HTML comments and use an alternative approach if it does please. Would a format like the following separated by blank lines read any better.

```

-# original header template
<their code>

-# custom service template
<our code>

```

app/assets/stylesheets/metadata.scss
```diff
@@ -201,4 +201,180 @@ dl.gem-c-metadata__list {
   padding: 15px 10px 15px 20px;
   margin-top: 60px;
   border-left: 5px solid #347ca9;
-}
\ No newline at end of file
+}
```
peterdavidhamilton, line 204: If the entire contents of this file is responsible for styling the `citation` partial, let's call it that. Anything that is unrelated should be moved out.

app/assets/stylesheets/links.scss
```diff
@@ -0,0 +1,9 @@
+.govuk-breadcrumbs li {
```
peterdavidhamilton, line 1: This file is specifically about the breadcrumb styling.

app/assets/stylesheets/other-resources.scss
```diff
@@ -0,0 +1,99 @@
+.hf-other {
+  padding-top: govuk-spacing(9);
+  padding-bottom: govuk-spacing(9);
+}
+
+.hf-ctas {
```
peterdavidhamilton, line 6: I see this class used only once in the CTA view partial.

app/assets/stylesheets/application.scss
```diff
@@ -25,24 +25,19 @@ $govuk-page-width: 1200px;
...
-@import '_cookie-banner';
-@import '_links';
-@import '_flash';
+@import 'print';
+@import 'card';
+@import 'side-nav';
+@import 'other-resources';
+@import 'cta';
 
+@import 'subnav';
```
peterdavidhamilton, line 39: This file has styling (that isn't imported) which can still be tidied up and removed. `app-sidebar` for example.

peterdavidhamilton (changes requested):

app/components/header_component.html.slim
martikat: This comment isn't shown when inspecting where it is used. I will update this so it is more readable though 👍

app/assets/stylesheets/subnav.scss
```diff
@@ -32,12 +32,12 @@
 
 .app-subnav--mobile {
```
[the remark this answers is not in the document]
martikat: It looks like this isn't being used. I will go ahead and remove this

app/assets/stylesheets/application.scss
martikat: I have removed subnav in the next commit

app/components/header_component.html.slim
```diff
@@ -6,11 +6,7 @@
         = image_tag 'dfe-logo-alt.png', class: 'dfe-logo-hover', alt: 'DfE homepage'
     #service-name
       = link_to service_name, service_url, class: 'dfe-header__link dfe-header__link--service'
-    #content-header.dfe-header__content.dfe-header-f-dtonly.dfe-header-f-toplinks
-      .dfe-header__menu
-        button#toggle-menu.dfe-header__menu-toggle aria-controls='header-navigation' aria-expanded='false' data-action='click->reveal#toggle'
-          | Menu
-    / This logic has been commented out as it is not required currently
+    / Original header template
```
peterdavidhamilton, line 9: `/!` this is the syntax I was concerned about to avoid making it visible the browser. I would recommend leveraging the single `/` where possible when it is all indented.

martikat: I have updated this in the next commit to use a single `/` for the comment

app/assets/stylesheets/cta.scss
```diff
@@ -15,6 +15,26 @@
...
+    padding: 20px;
+  }
+
+  .table-scroll {
+    overflow: scroll;
+  }
+}
+
 @media (max-width: 40.0525em) {
   .hf-card {
```
peterdavidhamilton, line 39: 👀

app/assets/stylesheets/downloadable-media.scss
```diff
@@ -0,0 +1,175 @@
...
+  -moz-osx-font-smoothing: grayscale;
+  font-weight: 400;
+  font-size: 16px;
+  font-size: 1rem;
+  line-height: 1.25;
+  margin: 0 0 15px;
+  color: #505a5f
+}
+
+@media print {
```
peterdavidhamilton, line 138: Please combine the `print` media block definitions.

peterdavidhamilton (changes requested):
This [npm package](https://github.com/alphagov/stylelint-config-gds) replaces [this rubygem](https://github.com/alphagov/scss-lint-govuk). Can you add it and see what it says?

app/assets/stylesheets/cta.scss
martikat: This can be deleted (it is not being used) 👍  I will include this change in the next commit

app/models/page.rb
peterdavidhamilton: I have addressed this in https://github.com/DFE-Digital/early-years-foundation-reform/pull/781

peterdavidhamilton (approved):

PR #822, opened 2024-06-28
ER-1158 DfE Frontend update

martikat:
[ER-1158](https://dfedigital.atlassian.net/browse/ER-1158)


app/components/header_component.html.slim
```diff
@@ -21,6 +19,8 @@
       .dfe-header__menu
         button#toggle-menu.dfe-header__menu-toggle aria-controls='header-navigation' aria-expanded='false' data-action='click->reveal#toggle'
           | Menu
+  = tag.div(class: 'dfe-width-container dfe-header__service-name')
```
peterdavidhamilton, line 22: Why use Ruby for the wrapper?

app/assets/stylesheets/application.scss
```diff
@@ -5,18 +5,12 @@
 // https://design.education.gov.uk/design-system/dfe-frontend
 // https://design-system.service.gov.uk/styles/colour/
 
-// https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_colours.scss
-
 
 $govuk-new-link-styles: true;
 $govuk-page-width: 1200px;
```
peterdavidhamilton, line 10: Is this needed?

package.json
```diff
@@ -3,16 +3,16 @@
   "packageManager": "yarn@4.0.2",
   "scripts": {
     "build": "esbuild app/javascript/*.* --bundle --sourcemap --outdir=app/assets/builds --public-path=/assets",
-    "build:css": "sass ./app/assets/stylesheets/application.scss:./app/assets/builds/application.css --no-source-map --load-path=node_modules --quiet-deps",
+    "build:css": "sass app/assets/stylesheets/application.scss app/assets/builds/application.css --no-source-map --load-path=node_modules --quiet-deps",
```
peterdavidhamilton, line 6: Has the CLI changed?

app/components/header_component.html.slim
martikat: I have updated this so it is just the div and class on this line

app/assets/stylesheets/application.scss
martikat: I have put this back in as it does have some useful information

package.json
martikat: This matches with how CDT has been setup/updated

peterdavidhamilton: The colon to space separation caught my eye.

peterdavidhamilton (changes requested):

peterdavidhamilton:
@martikat there are deprecation warnings due to assets we don't have. References to things we don't need or can't find should be removed rather than errors be silenced.
```
DEPRECATION WARNING: The asset "images/govuk-mask-icon.svg" is not present in the asset pipeline.
Falling back to an asset that may be in the public folder.
This behavior is deprecated and will be removed.
To bypass the asset pipeline and preserve this behavior,
use the `skip_pipeline: true` option.
 (called from index at /home/runner/work/early-years-foundation-reform/early-years-foundation-reform/app/controllers/pages_controller.rb:9)
```

peterdavidhamilton (changes requested):
In case you didn't compare with CDT, the last commit was an opportunity to use a partial so the head was consistent and future changes happen in one place. Could we copy that please.

peterdavidhamilton (approved):

PR #859, opened 2024-09-11
Remove HFEYP Beta Label

David-Feetenby:
Remove HFEYP Beta label by deleting references, removing the slim and changing the margin to padding


martikat (changes requested):
The `page_spec` needs to be updated here to pass. You will also need to update `page_footer_spec` when the feedback link has been added to the footer

martikat (approved):

PR #987, opened 2025-04-17
Make environment available on rails config bject and use to instantia…

benvand:
…te Sentry with an environment

What happens if the env var isn't there

```
config.environment = ENV['ENVIRONMENT']
```
??

Why, in the config do we do this in a number of ways? Which is best for this addition?
eg

```
ENV.fetch('CONTENTFUL_SPACE', credentials.dig(:contentful, :space))
ENV.fetch('FEEDBACK_URL', '#FEEDBACK_env_var_missing')
ENV['ENVIRONMENT']  # ENV['ENVIRONMENT'].eql?('development')
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,7 @@ on: [push]
 
 env:
   RAILS_ENV: test
+  ENVIRONMENT: test
 
 jobs:
   tests:
--- a/config/application.rb
+++ b/config/application.rb
@@ -63,8 +63,8 @@ module GovukRailsBoilerplate
     #
 
     # @return [Boolean]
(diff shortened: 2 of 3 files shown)


benvand:
https://docs.sentry.io/platforms/ruby/guides/rails/configuration/environments/

How to set envs up

https://github.com/search?q=%22Sentry.init+do+%7Cconfig%7C%22&type=code

We're using enabled_environments wrong
https://docs.sentry.io/platforms/ruby/guides/rails/configuration/options/#enabled_environments

martikat (approved):

benvand:
Looks ;like using fetch is best so we error out https://stackoverflow.com/questions/16569409/fetch-vs-when-working-with-hashes

PR #1165, opened 2026-06-23
eyp-264/redis-cache

david-dfe:
Add Redis support for caching; implement contentful resilience in Page and Resource models


config/environments/production.rb
```diff
@@ -57,8 +57,21 @@
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
 
-  # Use a different cache store in production.
-  # config.cache_store = :mem_cache_store
+  # Use Redis for cross-instance cache sharing when configured.
+  config.cache_store = if ENV['REDIS_URL'].present?
```
haridfe, line 61: Is URl add to environment variable ?

david-dfe: Redis_url is injected by as an app service setting in terraform which azure injects in to the container at runtime. It's built from the provisioned cache in terraform-azure/local.tf

That then flows through main/tf>web module>web_app.app_settings. It was added in #1170 The guard is intentional so that we use the redis store where the var is present but in local env where its unset we fall back to memory_store.

PR #1171, opened 2026-06-29
Add Redis private endpoint configuration and update related variables

david-dfe:


harkdfe (approved):
Configuration working as expected. Public access is disabled.

PR #1175, opened 2026-06-29
Switch Redis from retired Redis Enterprise to classic Azure Cache for redis

david-dfe:
- redis.tf: replace azapi Microsoft.Cache/redisEnterprise (Azure Managed Redis) cluster + database with azurerm_redis_cache. Redis Enterprise creation is retired and AMR Balanced SKUs were not provisionable in westeurope. Tier by environment: Standard C1 for dev, Premium P1 otherwise. Private endpoint subresource changed to redisCache.
- local.tf: REDIS_* app settings now read from azurerm_redis_cache attributes (.redis.cache.windows.net host, SSL port 6380, no username) in both webapp and slot blocks.
- terraform-azure-network/vnet.tf: private DNS zone changed to privatelink.redis.cache.windows.net (classic Redis zone).


harkdfe (approved):
good.

